PR #42, opened 2026-07-19
Discover publications by the people you follow on Bluesky

hipstersmoothie:
Readers sign in with Bluesky already following people who publish on standard.site — and the app never told them. On my own account that's **240 people I follow, writing 362 publications**, none of which were surfaced anywhere.

## Surfaces

1. **Discover entry point** — a "Find your friends" strip above Trending with an overlapping avatar stack and a live count. Renders only when there's someone to show, matching how the neighbouring rails behave.
2. **`/friends`** — the destination. Person heading (avatar, name, handle, Follow) with their publications in a quieter well beneath, each carrying its own Subscribe state. Paginated at 12 people with infinite scroll.
3. **Onboarding** — a new `friends` step between topics and follow, showing the top 6 with a "N more…" line. Selection stays in-memory like the rest of the wizard; people running several publications get a "Select all".

## How the join works

Candidate-first: the read-model already knows every DID that publishes, so we ask the Bluesky AppView *"of these, which do I follow?"* via `app.bsky.graph.getRelationships` rather than paginating the reader's follow list. Cost tracks the size of the author set instead of the reader's follow count, and nothing about the Bluesky graph is stored.

Results group by person, because that's the unit the reader thinks in — "Anna writes here", not "here is a publication that happens to be Anna's". Following a person also auto-subscribes their future publications, which is what the person-level action is for.

## Sizing came from measurement

The first working version was **20s load, 556KB payload, 45,795px tall, 11,698 DOM nodes** — and carried a correctness bug: a 900-DID cap meant to bound the AppView calls was silently hiding two thirds of the network's **2,754** publishing authors.

Benchmarking showed the cap was solving a problem that didn't exist — 30 batched `getRelationships` calls take **260ms**. So:

- batches now run **concurrently** (8 at a time) instead of serially
- cap raised to **5,000** — full network coverage, `truncated: false`
- response **paginated**, with the per-reader graph sweep cached server-side so page two is DB-only
- candidate list **memoized process-wide** (identical for every reader, cost ~900ms)
- follow-status caches **seeded from the payload**, so a page of ten people doesn't fire twenty status requests

Result: **3.1s cold, 399ms warm, 36KB, 1,620 nodes.**

## Edge cases

- A dedicated wizard step rendering nobody would be a dead screen, so the welcome loader prefetches the lookup and the flow **skips the step when it's empty**, progress dots adjusting to match. A deep link to `?step=friends` with nothing to show moves on by itself.
- An unreachable AppView and a failed request both report **"couldn't check"** — never "nobody you follow publishes here". Partial sweeps aren't cached.
- Signed-out, long handles/display names, people with one vs. many publications, and the truncation ceiling all handled.

## Drive-by fix

`OnboardingPubRow` descriptions were clipped mid-word on a single line. The row is a `<button>` (inheriting `white-space: nowrap`) and StyleX doesn't emit the `-webkit-box-orient` the line clamp depended on. **Pre-existing on the current follow step** — this repairs both.

## Verification

Screenshotted signed-in at desktop light, desktop dark, and 390px mobile, plus both onboarding steps. Typecheck clean, oxlint clean, oxfmt applied, **276 tests pass** (5 new, covering ranking, tie-breaking, identity fallback, and dropping followed accounts that publish nothing), production build passes, i18n catalogs extracted (30 new messages × 9 locales).

## Worth a look

- **"Follow" vs "Subscribed"** reads a little oddly on a person whose only publication you already subscribe to. They are genuinely different records and the masthead dek explains it, but the labels sit close together.
- Route is `/friends`, copy is "Find your friends" — easy to swap for `/following` if you'd rather.

### DIFF
--- a/src/components/onboarding/onboarding-pub-row.tsx
+++ b/src/components/onboarding/onboarding-pub-row.tsx
@@ -72,17 +72,19 @@ const styles = stylex.create({
   },
   description: {
     color: uiColor.text1,
-    display: "-webkit-box",
     fontFamily: fontFamily.sans,
     fontSize: fontSize.sm,
     lineHeight: lineHeight.sm,
     marginBottom: 0,
     marginTop: 0,
+    // The row is a <button>, which sets `white-space: nowrap` — inherited, that
+    // pinned descriptions to one unwrappable line that got clipped mid-word.
+    // Restore normal wrapping, then cap the block at two lines.
+    maxHeight: `calc(2 * ${lineHeight.sm} * ${fontSize.sm})`,
+    minWidth: 0,
     overflow: "hidden",
-    // eslint-disable-next-line @stylexjs/valid-styles
-    WebkitBoxOrient: "vertical",
-    // eslint-disable-next-line @stylexjs/valid-styles
-    WebkitLineClamp: 2,
+    overflowWrap: "anywhere",
+    whiteSpace: "normal",
   },
   metaLine: {
     alignItems: "center",
--- a/src/components/onboarding/step-friends.tsx
+++ b/src/components/onboarding/step-friends.tsx
@@ -1,0 +1,151 @@
+import { Plural, Trans } from "@lingui/react/macro";
+import * as stylex from "@stylexjs/stylex";
+import { useQuery } from "@tanstack/react-query";
+import { Check, Plus } from "lucide-react";
+
+import { discoverApi } from "#/integrations/tanstack-query/api-discover.functions";
+import type { FriendPublisher } from "#/integrations/tanstack-query/api-discover.functions";
+import { ONBOARDING_FRIENDS_LIMIT } from "#/lib/onboarding";
+
+import { Button } from "../../design-system/button";
+import { Flex } from "../../design-system/flex";
+import { uiColor } from "../../design-system/theme/color.stylex";
+import { verticalSpace } from "../../design-system/theme/semantic-spacing.stylex";
+import { spacing } from "../../design-system/theme/spacing.stylex";
+import {
+  fontFamily,
+  fontSize,
+  lineHeight,
+} from "../../design-system/theme/typography.stylex";
+import {
+  FriendPublisherGroup,
+  FriendPublishersDegradedNote,
+  FriendPublishersSkeleton,
+} from "../reader/friend-publishers";
+import { OnboardingPubRow } from "./onboarding-pub-row";
+
+const styles = stylex.create({
+  groups: {
+    display: "flex",
+    flexDirection: "column",
+    rowGap: verticalSpace["10xl"],
+  },
+  more: {
+    color: uiColor.text1,
+    fontFamily: fontFamily.sans,
+    fontSize: fontSize.sm,
+    lineHeight: lineHeight.sm,
+    marginBottom: spacing["0"],
+    marginTop: spacing["0"],
+  },
+});
+
+/**
+ * Onboarding step: publications written by the people the reader follows on
+ * Bluesky — the strongest signal available at first run, so it comes before the
+ * topic and trending suggestions.
+ *
+ * Selection is in-memory (committed with the rest of the wizard's picks at the
+ * end), which is why this uses {@link OnboardingPubRow} rather than the live
+ * subscribe buttons on `/friends`.
+ */
+export function StepFriends({
+  selected,
+  onToggle,
+}: {
+  selected: Set<string>;
+  onToggle: (uri: string, next: boolean) => void;
+}) {
+  const { data, isPending } = useQuery(
+    discoverApi.getFriendPublishersQueryOptions({
+      limit: ONBOARDING_FRIENDS_LIMIT,
+    }),
+  );
+
+  if (isPending) {
+    return <FriendPublishersSkeleton groups={2} />;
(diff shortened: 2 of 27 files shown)


hipstersmoothie:
Reworked per feedback, and rebased onto `main`.

**Dropped the per-person grouping.** The person-level Follow button sitting above a row already marked "Subscribed" was two similar affordances stacked, and the one on top often looked like it contradicted the one below. `/friends` and the onboarding step are now plain lists of publications using the same rows the Discover directory and tag pages already use — which name their author on every row, so the author survives and the ambiguity doesn't.

That deletes `FriendPublisherGroup`, the person-level follow control, the "Select all" shortcut, and the `hideOwner` prop this PR had added to `PubDirectoryRow` (so `cards.tsx` is back to baseline).

**Already-subscribed publications are filtered out** server-side — this is a page of things to *discover*, and rows pre-marked "Subscribed" are noise. On my account that's 362 → 299 publications, 240 → 206 writers.

**Rows you subscribe to during a visit stay put.** Subscribing invalidates the whole `["discover"]` query prefix, which would have made a row vanish the instant you subscribed to it. The friends queries moved to a `["friends"]` key namespace (plus `refetchOnWindowFocus: false`), so a row settles into "Subscribed" and only drops out on the next visit.

**Rebased onto main** to pick up the card overflow fixes from #38 — worth noting, since the layout complaint that started this was on a preview built before those landed.

Re-verified signed in at 390 / 768 / 1024 / 1280: no horizontal spill at any width (`documentElement.scrollWidth === innerWidth` throughout). Typecheck, oxlint, oxfmt, 277 tests, and the production build all pass. Tests were rewritten for the flat model and now also cover the subscribed-exclusion.

Note the previously-flagged "Follow vs Subscribed" open question is resolved by this change — there's only one affordance per row now.

hipstersmoothie:
Added the Publications / Articles tabs.

**Publications leads**, Articles is second — the page exists to get you subscribed, and the articles are what those publications happen to have published lately. Follows the tag pages' pattern exactly: URL-backed `view` search param, the shared `Tabs`/`TabList`/`TabPanel` components, and the inactive panel left unmounted so an unvisited tab costs no request.

**Both tabs describe the same set.** The Articles feed is scoped to the publications the Publications tab lists — so already-subscribed publications are excluded from both, and the two tabs never disagree about what "people you follow" means. Nothing you'd already get on Home is repeated here. The expensive Bluesky sweep is shared between tabs via the existing server-side graph cache, so switching tabs is a DB-only query.

**Fixed a hydration mismatch this feature introduced.** The friends data is prefetched off the critical path, so the server could finish that fetch mid-render and emit loaded markup while the client hydrated without it — React then threw away the tree. Both surfaces now render their pending branch until mount (new `useHasMounted` hook), which makes hydration deterministic. Verified clean over repeated loads on `/friends` in both tabs.

⚠️ **Pre-existing, not from this PR:** `/discover` still throws an intermittent hydration mismatch. It reproduces **signed-out**, where this feature's prompt cannot render at all, so it predates this work — flagging it rather than fixing it here since it's a separate race in the existing rails. Happy to open a follow-up issue.

Verified signed in at 1280 and 390. Typecheck, oxlint, oxfmt, 277 tests, production build, and i18n extraction all pass.

hipstersmoothie:
Three fixes pushed.

**1. Trending's limit select stays inline with the heading.** `SectionHead` already had a `stackOnMobile` prop for exactly this case — the select is a compact control, not a full-width mobile action. Verified on the same row at 390 and 360.

**2. Publication name and `@handle` share a line.** They stacked unconditionally below 40rem. Now inline, with the mobile name stepped down `xl` → `lg` so the pair fits the ~240px a phone row actually gives you, and the gap tightened.

One caveat worth stating plainly: **they still wrap when the name is long.** I tried forcing a single line and measured why it can't work — at 390px the text column is 242px, and e.g. `ATProto Community` (152px) + `@atmosphere.community` (145px) needs 305px. Forcing one line means truncating, and truncating either the publication name or the handle costs real information on a directory of publications you don't know yet. So: inline whenever it fits, wrap as the fallback. If you'd rather truncate the handle to guarantee one line, that's a one-line change — say the word.

This row is shared with Discover, tag and search. Checked all of them plus `/friends` at 360 and 390: no horizontal overflow anywhere.

**3. Discover blocks on the friends lookup, but only when it's cheap.** The loader races the prefetch against a 700ms budget, and the race *is* the warmth test — the graph is cached server-side for ten minutes and answers warm in ~400ms, while a cold sweep of the whole author set can't. Measured: warm loads now paint the prompt in the first response (TTFB ~670–850ms); a cold load loses the race and streams in as before, capped at 700ms of extra wait. Since the warm path settles the query before render, the prompt no longer needs its mount-gate.

`/friends` deliberately keeps streaming with a skeleton rather than blocking — a full-page list is better served that way.

**On the `/discover` hydration mismatch I flagged earlier:** I pinned it down while measuring. The diff points at `DiscoverRecommendedSection`'s `SectionHead`, not the friends prompt — same defect class (a non-blocking `getDiscoverExtras` prefetch resolving mid-render), but pre-existing code. I left it alone rather than change the timing of rails outside this PR's scope. Happy to fix it here or in a follow-up, your call.

PR #80, opened 2026-07-23
i18n: update Portuguese translation

victoriamdo:


src/locales/pt/messages.po
```diff
@@ -822,7 +822,7 @@ msgstr "Ainda não há publicações indexadas."
 #. placeholder {0}: result.added
 #: src/components/reader/subscriptions-table.tsx
 msgid "{0} added to {listName}."
-msgstr ""
+msgstr "{0, plural, one {# adicionado para } other {# adicionados para }} {listName}"
```
victoriamdo: Question, would that work? I added a variation based on the number.

hipstersmoothie: i tested and this does seem to work. I removed the extra space

PR #89, opened 2026-07-25
Sort + reorder the sidebar's flat subscription rows

hipstersmoothie:
Discussion: at://did:plc:esmiuxk53vmsllayghrq676w/app.userinput.discussion/3mrdmxtfqjsco

## Original request

> Title: Reorder subscriptions
>
> It would be neat to be able to reorder subscriptions in a list:
> - automatically, for instance alphabetically or by some count (such as unread)
> - manually

## What I found

- **The "automatic" half already exists.** `/subscriptions` (`subscriptions-table.tsx`) is a fully sortable table of every followed publication and person — Name (alphabetical), Unread, Last post, Articles, Followers, Topic, Lists are all sortable columns, ascending/descending. So "sort alphabetically or by unread count" was already shipped there.
- **The gap is the sidebar itself.** The "Subscriptions" section in the persistent left sidebar (and the mobile sheet) renders followed publications/people as flat rows in a fixed order (most-recently-posted publications first, then people by follow date) with **no way to sort or manually reorder them** — that's the actual thing the reporter ran into day to day, since that's the list they see on every visit, not the `/subscriptions` management page.
- There was already a very close precedent to build from: **"Reorder lists"** (`reorder-lists-modal.tsx`) — a drag-and-drop dialog for ordering the sidebar's *list groups* (folders), persisted on `app.standard-reader.sidebarPref`. It didn't cover the *flat, ungrouped* rows (subscriptions not filed into a list), which is what this PR adds.
- The reporter's suggested solutions (alphabetical / by-unread-count / manual) all map directly onto what got built — no divergence from what was asked, just scoped to where the actual gap was (the sidebar, not the already-sortable directory table, and not list-group membership order, which was also already solved).

## What changed

A **Sort** menu next to the existing New list / Reorder lists / Collapse-all icons in the sidebar's "Subscriptions" header, covering the ungrouped (flat) publications and people:

- **Recent activity** (default, unchanged) — the existing most-recent-first order.
- **A–Z** — by display name.
- **Most unread** — highest unread count first.
- **Custom order…** — opens a drag-and-drop reorder dialog (`ReorderSubscriptionsModal`), the same react-aria `useDragAndDrop` + `ListBox` interaction as the existing "Reorder lists" dialog, just listing subscriptions with avatar + name instead of list names.

The first three modes interleave publications and people into one combined ranking (`orderSubscriptions` in `use-sidebar-pref.ts`); "Recent activity" is a no-op that preserves today's pubs-then-people split exactly. The chosen mode and any manual order persist per-reader on `app.standard-reader.sidebarPref` (new `subscriptionSort` / `subscriptionOrder` fields, mirrored to `sidebar_prefs` via the tap ingester, migration `drizzle/0021_shallow_inhumans.sql`), following the exact same optimistic-cache-then-write-through pattern as the existing `listOrder`/`collapsed` fields.

Out of scope, deliberately: the `/subscriptions` directory table (already fully sortable) and the order of members *within* a list group (unrelated feature).

### Shape brief (craft)

This went through `/impeccable craft`. Since this run is unattended, the brief below was treated as confirmed once written (per this routine's rules), with the PR itself as the review gate:

> Add a small **Sort** control to the sidebar's "Subscriptions" section header (next to the existing New list / Reorder lists / Collapse-all icon buttons), governing the order of the *ungrouped* (flat) subscription rows. Modes: Recent activity (unchanged default), A–Z, Most unread, Custom order. Custom order opens a drag-and-drop reorder dialog, visually and behaviorally identical to the existing "Reorder lists" dialog. Persistence mirrors the existing `listOrder`/`collapsed` pattern on `sidebarPref`. Zero new visual vocabulary — reuse `IconButton`, `Menu`, `Dialog`, `ListBox`/`ListBoxItem`, the grip-drag pattern, and existing theme tokens; this is a Product-register control (earned familiarity), not a design moment. Out of scope: the `/subscriptions` table and ordering within a list group.

**Screenshots:** not included. This environment has no live Neon database or authenticated reader session to populate a real sidebar with subscriptions, so I couldn't drive a signed-in browser session to capture the feature in place. The dialog and menu are built entirely from existing, already-screenshotted design-system primitives (`Menu`/`MenuItem`, `Dialog`, `ListBox`/`ListBoxItem`, `IconButton`, `Avatar`) in the exact structural pattern as the shipped "Reorder lists" dialog, so the visual risk is low, but I want to flag the gap honestly rather than skip it silently.

## Testing

- `pnpm build && pnpm lint && pnpm typecheck && pnpm test` all pass locally.
- `pnpm exec drizzle-kit check` passes (migration journal/snapshot consistent).
- `pnpm i18n:extract` run and catalogs committed (new strings: Sort subscriptions, Recent activity, A–Z, Most unread, Custom order…, Reorder subscriptions dialog copy).
- Not run: `pnpm lex:lint` (the `goat` CLI isn't installed in this sandbox) — the lexicon change is additive-only (two new optional fields, no changes to existing ones), so this should be a safe assumption, but the `lexicons` CI job will confirm.

---
_Generated by [Claude Code](https://claude.ai/code/session_01QXBr7zyMzEjYDhychBhLcj)_


hipstersmoothie:
Good catch — pushed in 6bb7e4b.

**A–Z** and **Most unread** now also apply to list groups:
- Each group's own members (publications + people, combined the same way as the flat rows) get ranked by name / unread count.
- The groups themselves get reordered by name / by their total unread count.

**Recent activity** and **Custom order** stay no-ops for groups — they leave each list's own manual member order (`ListEditModal`'s drag-to-reorder) and the manual group order (`ReorderListsModal` / "Reorder lists") alone, since both already have dedicated reorder UI of their own. Only the "compute a full ranking" modes (alpha/unread) override those.

Applies to both the desktop sidebar and the mobile Subscriptions sheet. Build/lint/typecheck/test all green on the new commit.

---

hipstersmoothie:
Found it — pushed in 9732c9e (which also merges in `main`, since the PR had drifted into a real conflict on `drizzle/meta/*` after `main` picked up its own migration at `0021`; renumbered mine to `0022_melted_joseph.sql` and regenerated the snapshot).

The bug: the Sort button was gated on `hasUngrouped` (only shown when there were flat/ungrouped rows). Once the group-sorting extension landed, Sort also controls list-group order — so a reader whose subscriptions are entirely organized into list groups (no ungrouped rows) saw no Sort button at all. Fixed the gate to `hasUngrouped || hasListGroups`.

Build/lint/typecheck/test all green on the new commit; `drizzle-kit check` confirms the renumbered migration is consistent.

---

hipstersmoothie:
CI failure diagnosed and fixed in 62af538.

**Root cause:** `pnpm db:migrate` (drizzle-kit's migrator) doesn't track "already applied" by hashing SQL content — it compares each migration's journal `when` timestamp against the single most-recent `created_at` recorded in the target DB. When I renumbered `0021_shallow_inhumans.sql` → `0022_melted_joseph.sql` to resolve the collision with `main`'s own new `0021_melodic_texas_twister.sql`, I regenerated it via `db:generate`, which stamped a **fresh** timestamp instead of the one it was actually applied under on this PR's persistent Neon test branch (`pr-89`, reused across CI runs). Drizzle then thought the migration was new, tried to re-run its `ADD COLUMN subscription_sort/subscription_order` statements, and hit "column already exists" — the actual error was swallowed by drizzle-kit's spinner, which is why the log just showed a bare exit code 1.

**Fix:** restored the journal entry's `when` to `1784967467704` — the timestamp from the original (pre-rename) `0021_shallow_inhumans` entry, which is what's actually recorded as applied on that branch. Drizzle now correctly recognizes the SQL as already run and skips it; content/schema/`drizzle-kit check` are unaffected, only the bookkeeping timestamp changed.

One side effect worth flagging honestly: `main`'s `0021_melodic_texas_twister` (`user.appearance` column) predates my migration's original timestamp, so on this specific reused `pr-89` branch it'll also read as "already applied" and get skipped — meaning `user.appearance` won't exist on that one test branch. That's an artifact of this branch's stale history, not a real defect: a fresh branch (or the actual production deploy) applies both migrations in correct chronological order with no collision.

---

hipstersmoothie:
Rebased onto `main` (now at 5527257) and squashed into a single commit (7b8fbe9) — the branch previously carried a merge commit from resolving the earlier drizzle collision, and a plain commit-by-commit rebase would have replayed that same collision (plus silently dropped a fix that had only ever landed inside the merge commit itself). Squashing the full diff and rebasing that onto `main` avoided both problems cleanly.

Verified before pushing: `pnpm build/lint/typecheck/test` all pass, `drizzle-kit check` is clean, i18n catalogs are up to date, and the Sort-button-visibility fix + migration-timestamp fix are both present in the squashed diff.

---

hipstersmoothie:
Confirmed fixed — CI is fully green on the fresh `pr-89` Neon branch (empty commit c84d8f7), including the migration-apply step. That was the last blocker; PR is ready.

---

hipstersmoothie:
Pushed 672d663 covering the last three rounds of feedback:

**Consolidated the crowded icon row.** Sort / Reorder lists / New list / Collapse all are now one overflow menu (`⋮`) next to "Subscriptions" instead of four separate icon buttons.

**Removed "Custom order" as a sort mode.** It's now its own direct action — **Reorder subscriptions…** — sitting alongside **Reorder lists…** in the menu, the same way those two already relate to each other, instead of competing with the automatic sort modes for a slot in that list. Dragging and saving still switches `subscriptionSort` to `"manual"` under the hood.

**Dropped the ascending/descending idea** per your follow-up — kept to three sorts that each mean something on their own (Recent activity / A–Z / Most unread) rather than doubling every entry with a reverse variant.

**Fixed the "PEOPLE" list bug you screenshotted.** Reproduced it exactly: for a list group, "Recent activity" was a no-op — but unlike the flat sidebar rows (which the server pre-sorts by real recency), a list's own membership array is just raw stored order, never recency-ordered. So the no-op showed arbitrary order, which is what "tay.zone" sitting first was — not a comparator bug (I verified the actual sort function directly against your exact name list and it sorts correctly). "Recent" (and "manual", which has no per-group equivalent) now fall back to A–Z for group content instead of leaving that arbitrary order in place.

Build/lint/typecheck/test all green locally; watching CI now.

---

hipstersmoothie:
Pushed a fix for the "recent sort doesn't work the same for both" issue: added a real **Default** sort mode (natural/stored order, untouched — now the actual default), and made **Recent activity** a genuine interleaved sort instead of the old "pubs-block, then people-block" no-op.

- **Default** — no automatic sort at all; preserves existing order for both the flat list and every list group's own membership/arrangement.
- **Recent activity** — now truly ranks publications (`lastDocumentAt`) and people (new `followedAt` field) together by actual recency, for both the flat rows and list groups.
- List groups now only auto-resort on an explicit **Recent activity** / **A–Z** / **Most unread** choice. **Default** and **Manual** leave a list's stored member order and your manual group arrangement alone.
- Since an active automatic sort would otherwise silently override a manual group arrangement, **Reorder lists…** is now disabled unless Sort is set to **Default**.

Includes a new migration (`0023_mushy_rawhide_kid.sql`) that flips the `sidebar_prefs.subscription_sort` default from `'recent'` to `'default'` and backfills any existing `'recent'` rows (none of this has shipped yet, so that's just test-branch cleanup).

---

hipstersmoothie:
Rebased onto `main`'s latest tip (it had advanced with its own new migration, `0022_silent_alex_power.sql`, colliding with this PR's `0022`). Also swapped the overflow menu trigger from an ellipsis to a cog icon per your feedback.

One heads up: resolving the migration collision meant regenerating this PR's `sidebarPrefs` migration fresh (now `0023_nosy_maverick.sql`, with a current timestamp so it's guaranteed to run after everything already on `main` — otherwise it would've silently been skipped in production for being timestamped earlier than your `0022`). That means the migration this PR ships no longer lines up with what's already been applied on the `pr-89` Neon test branch from earlier runs this session. Could you delete the `pr-89` Neon branch once more so CI provisions a clean one? I don't have Neon API access in this sandbox to do it myself.

Now starting on the bigger ask — turning "Reorder lists" into a full one-level tree editor (reorder lists, reorder/move pubs+people between lists, move them in/out of lists, interleave with list order) using react-aria-components' built-in drag-and-drop.

---

hipstersmoothie:
Pushed the full drag-and-drop tree. The desktop sidebar's Subscriptions section is now one level-deep tree (`design-system/tree`'s headless `Tree`/`TreeItem` + react-aria's `useDragAndDrop`, no new visual vocabulary) instead of a separate flat list + list-group sections — list groups and ungrouped publications/people are top-level siblings, and each list's own members are its children.

When Sort is **Default**, dragging supports everything you asked for:
- reorder lists
- reorder members within a list
- move a member between lists
- move a member into/out of a list
- reorder members relative to lists at the top level

An automatic sort (Recent activity/A–Z/Most unread) disables dragging, same reasoning as before — it computes its own arrangement, so a drag wouldn't stick.

Persistence: a new `treeOrder` field on `sidebarPref` captures the manual top-level arrangement (list at-uris interleaved with ungrouped ids); a new `setListMembers` mutation does a full publications/users replace in one write for a list's own membership/order. The old `ReorderListsModal`/`ReorderSubscriptionsModal` dialogs are gone, superseded by the inline drag.

Two scoping notes, called out honestly rather than silently:
- **Cross-kind order within one list isn't separately persisted** — dragging a publication next to a person inside the same list reorders correctly on the screen, but on save each kind (publications/people) keeps only its own relative order, same limitation `ListEditModal`'s member editor already has. True interleaving would need a new per-list `memberOrder` field; happy to add it if it matters in practice.
- **Saved lists (not yours) are read-only containers** — you can move their position among your top-level lists, but not drag their members (can't edit someone else's list).
- Mobile `SubscriptionsSheet` is unchanged (still read-only, no drag) — this pass is scoped to the desktop sidebar per your ask.

Couldn't visually browser-test the drag interaction itself — this sandbox has no live Neon DB or authenticated session (same constraint as this PR's original body). Build/lint/typecheck/test all pass and I read through the drag logic carefully, but flagging this gap rather than skipping it silently.

---

hipstersmoothie:
Addressed all four:

1. Swapped `Cog` (12-spoke, busy) for lucide's `Settings` icon (single rounded gear) on the overflow menu trigger.
2. Rebuilt the tree on raw react-aria-components `Tree`/`TreeItem`/`TreeItemContent` instead of `design-system/tree` — that wrapper baked in its own level-based indent/chevron-spacer/checkbox styling, which was pushing nested list members in from the edge. The new custom row markup keeps every row (top-level or nested) at the same flush left edge, matching the sidebar's original look.
3. Added a custom drag preview — a small pill showing the dragged item's name (+ a "+N" badge for multi-drag), same treatment the old "Reorder lists" dialog used.
4. Added a drop-target indicator — a thin line between rows showing exactly where the drop will land, via `useDragAndDrop`'s `renderDropIndicator`.

---

hipstersmoothie:
Pushed fixes for the four items:

1. **Reordering only "on" when explicit** — root-caused: `__root.tsx`'s root loader seeded `sidebar`/`lists`/`savedLists` from the bootstrap but never `sidebarPref`, even though `loadShellSnapshot` already returns it alongside the other three. That meant `subscriptionSort` (and collapse/hiddenNav) briefly read from the `EMPTY` fallback on every signed-in page load before a client-side fetch resolved the real value — so drag-and-drop could appear briefly enabled even when your saved sort wasn't "Default". Fixed by seeding `sidebarPref` in the bootstrap alongside the other three.
2. **Collapse state + hidden tabs applied on the server** — same root cause and same fix. List groups no longer render expanded-then-collapse and hidden nav items no longer flash visible-then-hidden; both are correct on first paint now since `sidebarPref` is genuinely in the SSR-seeded cache before the tree ever renders.
3. **Drag handles only show when sort mode is on** — this was already gated correctly via `allowsDragging` (verified against react-aria's source — it does reflect `useDragAndDrop`'s `isDisabled`), but was affected by the same flash: with the seeding bug fixed, handles now only ever appear when `subscriptionSort` is genuinely `"default"`.
4. **Chevron position** — moved it back to the end of the row (after the unread badge), matching the original layout. It had drifted to the start during the raw-react-aria-components rebuild.

Also hardened `ensureShellSnapshot`'s cache guard so a partial seed (sidebar without sidebarPref) can't silently skip the fallback fetch again in the future.

---

hipstersmoothie:
Pushed both:

1. **Drag-and-drop now requires an explicit toggle.** Added a **"Reorder subscriptions…"** menu item (a local, unpersisted UI state, not a `subscriptionSort` value) that the reader has to explicitly turn on before dragging activates — it relabels to **"Done reordering"** while active, and is disabled (and auto-turned back off) whenever an automatic sort is selected, since that computes its own arrangement. Previously drag was implicitly live any time sort happened to be "Default", including for readers who'd never touched Sort at all.
2. **Nested-row indent fixed.** Found the actual cause: the group header row used a smaller `columnGap` than a member row, so after the (same-width) drag handle, a group's name and a nested member's avatar landed at slightly different x-positions — not a real "indent" from the tree library (react-aria-components adds none), just a gap mismatch in my own row styles. Unified both to the same gap; a group's name and its members' avatars now start at the exact same edge.

While in there I also fixed a related bug: react-aria's default `hasChildItems` heuristic only counts a node has children if it has *more than one*, which would've hidden the expand chevron for any list with exactly one member. Now passed explicitly since a list always has at least one child (a real member or the "Empty list." placeholder).

---

hipstersmoothie:
Pushed fixes for the latest round of feedback:

1. **Drag handles were showing on every row regardless of reorder mode.** Root cause: `react-aria-components`' per-item `allowsDragging` render prop reflects only whether `useDragAndDrop` hooks exist at all (`!!dragState`), not whether they're actually enabled (`isDisabled`) — so it was structurally incapable of respecting the reorder-mode toggle. Fixed by computing a local `dragEnabled` flag (`reorderMode && subscriptionSort === "default"`) and using that everywhere a row decides whether to render its grip handle, instead of trusting the library's prop.
2. **Drop-target indication for lists.** A list row now gets a highlighted background + outline while it's the active "drop into" target, using the `data-drop-target` attribute react-aria-components sets on the target `TreeItem`.
3. **Drag handle icon alignment.** The grip icon was centered inside its button's hit-target box, which sat it slightly right of the row's true left edge (and out of alignment with the "SUBSCRIPTIONS" header). Changed to flush-align the icon to the button's own edge while keeping the same comfortable tap target size.
4. **Settings icon → checkmark while reordering.** The overflow trigger now swaps from the settings gear to a checkmark while reorder mode is active; pressing it directly exits reorder mode instead of opening the menu.

Verified with the full local gate (typecheck, lint, i18n:extract, format, build, test — all clean/passing).


---

hipstersmoothie:
Fixed both:

1. **Row height jumping in reorder mode.** The drag handle button was sized like the header's "sm" icon button (28px square) — taller than a row's avatar (24px) — so every row grew a few pixels the moment reorder mode turned on. Resized the handle to match the avatar's height exactly, so toggling reorder mode never changes row height.
2. **Excess spacing between the drag icon and avatar/label.** The handle's box was also wider than it needed to be, leaving empty trailing space before the column gap kicked in. Narrowed it so the gap now reads as intentional spacing, not a leftover button hitbox.

Full local gate (typecheck, lint, format, build, test) passing.


---

hipstersmoothie:
Pushed mobile parity for the subscriptions tree — the sort + reorder feature previously only existed on the desktop sidebar.

**What changed:**
- Extracted the tree's data-building (`useSubscriptionsTree`) and rendering (`SubscriptionsTree`) out of `app-shell.tsx` into two shared modules (`subscriptions-tree.tsx`, `use-subscriptions-tree.tsx`), so desktop and mobile use one implementation instead of two that could drift apart. The desktop sidebar computes the tree once and passes the same `topNodes`/`groupNodes`/`dragAndDropHooks` down to the mobile sheet, which mounts its own `<Tree>` from that identical data — react-aria's `dragAndDropHooks` is a stateless config bag, so it's safe for two separate `<Tree>` instances to share one.
- The mobile `SubscriptionsSheet` drawer now has the same overflow menu as desktop (Sort submenu, **Reorder subscriptions…**/**Done reordering** toggle with the settings-gear/checkmark swap, collapse/expand all) and the same drag-and-drop tree (custom drag preview, drop-line indicator, drop-target highlight).
- `reorderMode` is one shared, unpersisted toggle in `AppShell` rather than a separate copy per surface, since only one of the two tree mounts is ever visible at a given viewport width.
- Removed the old accordion-style `SheetPubRow`/`SheetUserRow`/`SheetListGroup` rendering and a dead, never-wired `onReorder` prop left over from the old reorder-lists dialog.

Verified with the full local gate (typecheck, lint, i18n:extract, format, build, test — all clean). One caveat: this sandbox has no `DATABASE_URL` configured, so I could only smoke-test that the app boots without a database; I wasn't able to click through the actual mobile drawer/drag flow in a signed-in browser session here — worth a manual pass on your end before merge.


---

hipstersmoothie:
Pushed two more changes:

1. **Moved the mobile settings icon into the action row.** It now sits alongside Add publication / New list as a large icon button (matching their height) instead of a smaller control floating in its own row above the tree.
2. **Squashed the sidebarPref migrations into one.** `0023` (added `subscription_sort` + a `subscription_order` column), `0024` (dropped `subscription_order`), and `0025` (added `tree_order`) were split into three because generating the drop+add as one migration hit drizzle-kit's interactive rename-detection prompt mid-development. None of the three had reached `main`, so I squashed them into a single `drizzle/0023_zippy_boom_boom.sql` that just adds `subscription_sort` and `tree_order` directly — the transient `subscription_order` column never needs to exist. `drizzle-kit check` confirms the journal/snapshots are consistent, and it regenerated cleanly with no rename prompt (there's no column to disambiguate against once the transient one is gone).

⚠️ **One thing worth knowing about the migration squash:** the persistent Neon `pr-89` branch CI reuses across pushes to this PR already has the *old* `0023`/`0024`/`0025` applied (and thus already has both `subscription_sort` and `tree_order` — the same end state). Since the new single migration has a different file hash, drizzle-kit will treat it as unapplied and try to run it against that branch, which will fail with "column already exists" since the columns are already there from the old migrations. I don't have Neon API access from this sandbox to delete/reset that branch myself — if the next CI run's "Apply migrations to Neon branch" step fails with that error, deleting the `pr-89` Neon branch (Neon console, or your API key) should let CI recreate it fresh on the following push, applying the new single migration cleanly from scratch.

Full local gate (typecheck, lint, format, `drizzle-kit check`, build, test) passing.


---

hipstersmoothie:
Fixed the migration check without needing a Neon branch reset. Made the squashed migration idempotent with `IF NOT EXISTS` on both `ADD COLUMN`s, so it's a no-op on the persistent `pr-89` branch (which already has both columns from the old pre-squash migrations) and still applies cleanly on a genuinely fresh database.

I didn't just trust `drizzle-kit check` here (it only validates journal/snapshot consistency, not that the SQL actually runs) — spun up a real local Postgres and ran the migration against both an empty `sidebar_prefs` table and one pre-populated with `tree_order`/`subscription_sort` already present. Succeeds in both cases (`NOTICE: column ... already exists, skipping` on the pre-populated one, exit 0).

Full local gate (typecheck, lint, format, `drizzle-kit check`, build, test) passing.


---

hipstersmoothie:
Fixed. When the subscriptions tree was rebuilt on `react-aria-components`, a list group's own unread total picked up the same pill-badge (`UnreadBadge`) styling as individual publication/person rows. On the original sidebar, that badge was only ever used for members — a list's own header count was always a bare, unstyled `<span>` inheriting the group header's small uppercase label text. Reverted just that spot (list group header count); member row badges (nested inside a list and top-level ungrouped rows) are unchanged.

Full local gate (typecheck, lint, format, build, test) passing.


---

hipstersmoothie:
Fixed. The plain-text count still sat inside the row's own `columnGap` (same distance from the chevron as from the group name). The original sidebar grouped the count and chevron into their own sub-row with no gap between them — wrapped them the same way here so they sit flush, matching the original spacing exactly.

Full local gate (typecheck, lint, format, i18n:extract, build, test) passing.


---

PR #97, opened 2026-07-27
Add article sort controls to tag page Articles tab

hipstersmoothie:
## Summary

Adds a sort dropdown to the tag page's Articles tab, allowing users to rank articles by **Recent** (newest first, default), **Trending** (precomputed trending score), or **Most popular** (all-time likes + Bluesky backlinks). The sort preference is persisted in the URL search params independently from the Publications tab's directory sort.

## Key Changes

- **Frontend (tag page route)**
  - Added `articleSort` search param (enum: `recent` | `trending` | `popular`) to the tag search schema, separate from the Publications tab's `sort` param so each tab maintains independent state
  - Integrated `ArrowDownWideNarrow` icon and new `Select` component from design-system
  - Created `ARTICLE_SORT_OPTIONS` constant with localized labels
  - Added `onArticleSortChange` handler to update URL on sort selection
  - Styled the sort dropdown in the tab bar using flexbox layout (`tabBarSort` style) to sit on the trailing edge and wrap gracefully on narrow screens
  - Updated `TagArticlesPanel` to accept and pass `articleSort` to API queries
  - Updated route loader and prefetch logic to include `articleSort` in dependencies

- **Backend (query layer)**
  - Exported new `ArticleCardSort` type (`"recent" | "trending" | "popular"`)
  - Added `sort` field to `ArticleCardQuery` interface with documentation explaining each ranking strategy
  - Implemented `articleCardOrderBy()` function that applies the appropriate `ORDER BY` clause:
    - `recent`: newest-first (existing behavior)
    - `trending`: by `documents.trending_score` desc, then newest-first
    - `popular`: by `documentRecommendCountSql()` (all-time likes), then backlink count, then newest-first
  - Extracted `documentRecommendCountSql()` helper (scalar subquery for like counts) so it can be reused in both SELECT and ORDER BY without drift
  - Updated `selectArticleCards()` to use `articleCardOrderBy()` instead of hardcoded `documentsNewestFirst()`

- **API layer**
  - Added `articleSortEnum` validation and `TagArticleSort` type export
  - Updated `articlesPageInput` and `tagPageInput` schemas to include `articleSort` with default `"recent"`
  - Updated `getArticles` and `getTagPage` server functions to accept and pass `articleSort` to the query layer
  - Updated query option builders to include `articleSort` in the query key and function calls

- **Localization**
  - Added new message keys: `"Sort articles"`, `"Recent"`, `"Most popular"` across all locale files (en, ar, de, en-XA, es, fr, ja, ko, pt, zh)

- **Documentation**
  - Updated `APP_VISION.md` to describe the Articles tab sort feature and ranking strategies
  - Checked off completed task in `TODO.md`

## Implementation Details

- **URL state management:** The `articleSort` param is independent from `sort` (Publications tab), allowing users to switch between tabs without losing their sort preference on each
- **Ranking strategy:** Trending and Popular sorts rank rather than filter — score ties fall back to newest-first, keeping the list paginable on low-traffic tags
- **Performance:** Popular sort uses a correlated subquery (`documentRecommendCountSql`) which is acceptable for filtered lists (tag-scoped queries use indexed lookups) but would be expensive on unfiltered feeds
- **Layout:** The sort dropdown uses flexbox with `flexShrink: 0` and `marginInlineStart: auto` to stay on the trailing edge of the tab bar, wrapping to its own line on narrow screens

https://claude.ai/code/session_01Lh8FR6RwYzm4PLLA8RuHPM

### DIFF
--- a/APP_VISION.md
+++ b/APP_VISION.md
@@ -113,8 +113,12 @@ Sections, top to bottom:
 
 - Route `/tag/$tag`; linked from topic chips on article cards, publication cards, and
   article kickers.
-- **Articles** tab: chronological feed of indexed, published articles carrying the tag
-  (case-insensitive) on discover-eligible publications.
+- **Articles** tab: indexed, published articles carrying the tag (case-insensitive) on
+  discover-eligible publications, with a **Recent / Trending / Most popular** sort select
+  sitting in the tab row. Trending ranks by the precomputed `documents.trending_score`;
+  Most popular ranks by all-time likes then Bluesky backlinks. Both rank rather than
+  filter — score ties fall back to newest-first, so the list stays paginable on a
+  low-traffic tag.
 - **Publications** tab: discover-eligible publications with at least one such document,
   with per-publication **tagged-post counts**, Most posts / Readers / Active / A–Z sort,
   and grid ⇄ list toggle.
--- a/TODO.md
+++ b/TODO.md
@@ -366,6 +366,9 @@ in structured o11y (`observe`) and reads from the Neon read-model.
 - [x] Tag directory: `/tag/$tag` lists publications with indexed posts carrying the tag
       and per-publication tagged-post counts; topic chips on cards link there.
       `tagApi` in `api-tag.functions.ts`.
+- [x] Tag Articles tab sort: Recent / Trending / Most popular select in the tab row
+      (`?articleSort=`, its own search param so the Publications-tab `sort` is
+      independent). Backed by `ArticleCardSort` on `selectArticleCards`.
 - [x] Publication profile query (header + owner identity, recent writing,
       readers-also-follow). `publicationApi.getPublicationProfile`.
 - [x] Article query (full content + publication card + byline contributors +
--- a/src/integrations/tanstack-query/api-shapes.ts
+++ b/src/integrations/tanstack-query/api-shapes.ts
@@ -211,6 +211,26 @@ export function publicationCardColumns(schema: Schema) {
 }
 
 /**
+ * Scalar subquery: all-time non-deleted likes on a document. Served as an
+ * index-only scan by `recommends_doc_recommender_created_idx` (partial on
+ * `deleted = false`), so it is cheap enough to also drive an `ORDER BY` over a
+ * filtered document set (see the tag feed's "Most liked" sort).
+ *
+ * Shared with {@link articleCardColumns} so a query that both selects and sorts
+ * on the count uses one expression rather than two that can drift apart.
+ */
+export function documentRecommendCountSql(schema: Schema) {
+  const d = schema.documents;
+  const rec = schema.recommends;
+  return sql<number>`coalesce((
+      select count(*)::int
+      from ${rec}
+      where ${rec.documentUri} = ${d.uri}
+        and ${rec.deleted} = false
+    ), 0)`.mapWith(Number);
+}
+
+/**
  * Select-columns for an {@link ArticleCard}. Pulls the document plus its
  * publication's name/icon and the owner profile's banner; queries should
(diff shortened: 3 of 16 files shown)


hipstersmoothie:
CI is red only on the Railway `ingest` preview deploy — **"Deployment cancelled"**. I don't think this diff caused it, and there's no code change to push for it.

Evidence:

- All five GitHub Actions checks are green, including `Lint, format, typecheck, build`.
- Blast radius: `src/server/ingest/service.ts` and its transitive modules never import `src/server/reader/queries.ts`. The single touchpoint with `api-shapes.ts` is `sync.server.ts`'s `import type { Db, Schema }` — type-only, erased at compile, and neither type changed.
- Five other Railway services built and deployed from this exact SHA: `claudeslop`, `recompute-cron`, `botlabeler`, `digest-cron`, `thread-cron` all green (`web` still deploying). `recompute-cron` builds from the same repo root as `ingest`.
- "Deployment cancelled" is Railway's superseded/aborted status. A real problem in this diff would surface as `FAILED` (build error) or `CRASHED` (startup error), not cancelled.

Likely mechanism, **unconfirmed**: the `Point Railway preview at Neon branch` job completed at 04:31:04 and rewrites `DATABASE_URL` on the preview services; a variable change starts a fresh deploy and cancels the in-flight one. `ingest` was seemingly still building when that landed. I couldn't verify this — the Railway CLI isn't installed in this session and there's no project token, so I have no read access to the deployment or its logs.

**Needs a hand:** a redeploy of `ingest` in the PR-97 preview environment should clear it. I'm not going to trigger that myself — mutating a deploy is outside what I'd do without a nod, and I lack the access regardless. If it reproduces on a redeploy, that would point at something real and I'll dig in.

Unrelated to this status, one open item I flagged when I pushed: the new `popular` sort orders on a correlated subquery, so it's evaluated for every row the tag filter admits, not just the page. It's bounded by `documents_tags_norm_idx` with each count an index-only scan on the partial recommends index, but CI runs against an empty Neon branch and never exercises it. Worth an `EXPLAIN` against prod-scale data on the busiest tag (`pnpm perf:explain` is set up for this) before trusting its latency.

---

hipstersmoothie:
Update — **`web` self-resolved, and it confirms the mechanism.** Disregard the redeploy ask in my previous comment for that service.

`web` fired a "Deployment cancelled" status for deployment `72f4c8bd`, then went green four minutes later on deployment `a03b18a8` — the *original* one that had been pending since 04:31. So Railway created a duplicate deployment for the same service/SHA, cancelled the redundant one, and the original finished fine. The cancelled duplicate posts its own commit status, which transiently reddens the PR until the real deployment's success status lands.

That's the supersede/cancel pattern I guessed at, now confirmed on `web` — and it means "Deployment cancelled" here is a duplicate being reaped, not a deploy that failed.

Current state, 6 of 7 green:

| Service | State |
| --- | --- |
| `web` | ✅ Success — web-standard-reader-pr-97.up.railway.app |
| `claudeslop` | ✅ Success |
| `botlabeler` | ✅ Success |
| `recompute-cron` | ✅ Success |
| `digest-cron` | ✅ Success |
| `thread-cron` | ✅ Success |
| `ingest` | ❌ Deployment cancelled |

All five GitHub Actions checks remain green.

`ingest` is the last red. By the `web` precedent I'd expect a surviving deployment to report success on its own, but `ingest` has now gone longer without one than `web` took to recover, so its real deployment may never have been created. Still nothing to fix in this diff — the blast-radius argument from my previous comment is unchanged. I'll keep watching and will flag it if it doesn't clear; a redeploy of `ingest` alone would settle it if you want it green sooner.

---

PR #109, opened 2026-07-29
Fix long article links overflowing the reading column and widening the page

hipstersmoothie:
Discussion: at://did:plc:6ayddqghxhciedbaofoxkcbs/app.userinput.discussion/3mrsoirw7e2g2

## Original request

> Title: Failed test wrap
>
> i observed this text wrap bug which caused the overall zoomable area to be larger than intended.

The report came with a screenshot showing a link inside an article body making the page/viewport wider than intended — not visible from the payload's text alone, but confirmed once a maintainer pointed it out while I was investigating.

## What I found

My first pass misread "zoomable area" as the magazine reader's paginated canvas and fixed a real (but unrelated) wrap gap there — see "Also included" below. The actual report is about the **ordinary article reading view**: a link renders inside `articleBodyStyles.body` (`src/components/reader/content/body-styles.ts`), and that container never set `overflow-wrap`.

`article-view.tsx`'s `root` style deliberately carries **no outer `overflow-x` clip**, with a comment explaining why: every wide block — code, tables, math, image carousels — already scrolls inside its own `overflow-x: auto` box, so nothing needs an outer clip. A plain link isn't one of those boxed elements. When link text is a long, space-free run (a bare-URL autolink, or link text that mirrors its own href — both common with pasted URLs), it has no wrap opportunity and overflows the reading column with nothing to catch it, widening the whole page — and, on mobile, the viewport a reader can pinch-zoom/pan into.

This exact bug was already found and fixed once, for comments: `commentaryParagraph` in `src/components/reader/comments/comments-styles.ts` sets `overflowWrap: "anywhere"` with the comment *"Comment bodies routinely contain bare at:// URIs and long URLs with no break opportunity; without this they overflow the card and the viewport."* That fix was never carried over to the main article body, which is the gap this report hit.

### Fix

```diff
 body: {
   ...
   minWidth: 0,
+  overflowWrap: "anywhere",
 },
```

One property, on the container every article paragraph/heading/link inherits from — matching `commentaryParagraph`'s existing fix exactly.

### Verification (headless-browser repro against the real compiled app CSS)

Rendered a paragraph with a long unbroken link inside a 480px column, with and without the fix's atomic class:

- **Before**: paragraph `scrollWidth` = 1346px (866px past its 480px column); the link visibly runs off the page, and the document's `scrollWidth` balloons past the viewport.
- **After**: paragraph `scrollWidth` = 432px — wraps cleanly inside the column, no overflow.

## Also included: a second, unrelated wrap gap (magazine colophon)

While investigating I found and fixed a second, real bug in the same problem space, in the **magazine reader**'s end-card colophon page: `.mag .colophon-spread` was missing the `overflow-wrap: anywhere; max-width: 100%; min-width: 0;` protection every sibling flow section (`.editorial-spread`, `.feature-body`) already has, so a long unbroken token there could inflate `flow.scrollWidth` (which `readFlowMeasure()` uses to compute the magazine's page count) and produce phantom blank pages. It's a legitimate fix — verified the same way, before/after `scrollWidth` measurements — but it isn't what this report was about, so it's included as a secondary fix rather than the headline one.

```diff
 .mag .colophon-spread {
   justify-content: center;
+  max-width: 100%;
   min-height: 0;
+  min-width: 0;
+  overflow-wrap: anywhere;
   overflow-y: auto;
   -webkit-overflow-scrolling: touch;
 }
```

## Verification

- `pnpm build` ✅
- `pnpm lint` ✅ (0 errors)
- `pnpm typecheck` ✅
- `pnpm test` ✅ (535 passed, 4 pre-existing skips)
- `pnpm format:check` ✅ (also fixed two pre-existing, unrelated formatting issues and a stale i18n catalog that were already failing on the branch before this PR and blocking CI's combined lint/format/typecheck/build job)

### DIFF
--- a/apps/standard-reader/src/components/reader/content/body-styles.ts
+++ b/apps/standard-reader/src/components/reader/content/body-styles.ts
@@ -70,6 +70,16 @@ export const articleBodyStyles = stylex.create({
     lineHeight: READING_LINE_HEIGHT,
     marginTop: spacing["9"],
     minWidth: 0,
+    // The article scroller deliberately carries no outer `overflow-x` clip
+    // (see the comment on `article-view.tsx`'s `root` style) because every
+    // wide block — code, tables, math, carousels — already scrolls inside its
+    // own `overflow-x: auto` box. A plain link's visible text isn't boxed
+    // like that, so a long unbroken URL rendered as link text (a bare-URL
+    // autolink, or link text that mirrors its href) had no wrap opportunity
+    // and overflowed the reading column, widening the whole page and the
+    // viewport a reader can pinch-zoom into. Same fix, same reasoning, as
+    // `commentaryParagraph` in comments-styles.ts.
+    overflowWrap: "anywhere",
   },
   bodyAfterHero: {
     marginTop: spacing["0"],
--- a/apps/standard-reader/src/locales/ar/messages.po
+++ b/apps/standard-reader/src/locales/ar/messages.po
@@ -123,12 +123,12 @@ msgid "Every renderer takes one input for the document — the content payload o
 msgstr ""
 
 #: src/components/reader/about-view.tsx
-msgid "One tap from the reading view, on any device, instantly in sync. Your library is always where you left it."
-msgstr "نقرة واحدة من وضع القراءة، على أي جهاز، ومزامنة فورية. مكتبتك دائمًا حيث تركتها."
+#~ msgid "One tap from the reading view, on any device, instantly in sync. Your library is always where you left it."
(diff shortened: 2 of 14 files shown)


hipstersmoothie:
The required CI check (**Lint, format, typecheck, build**) is green. The Railway preview build failure on **claudeslop** is unrelated to this PR: `services/claudeslop`'s Docker build context never copies `apps/standard-reader/` (source, package.json, or locales) — nothing in this diff (magazine CSS, two formatting fixes, regenerated i18n catalogs) can reach that service's build. Not investigating further per the "infra failure, not your code" guidance; flagging here in case it needs separate attention.

---

PR #116, opened 2026-07-30
Fix Leaflet linkblog bodies collapsing markdown into one clump of text

hipstersmoothie:
Discussion: at://did:plc:fip3nyk6tjo3senpq4ei2cxw/app.userinput.discussion/3mruxyev4mcb3

## Original request

> Title: Body text from Skyreader's Linkblogs is not including new lines
>
> What I mean is that all text is rendered as if it was all just one big clump of text, when the data contains multiple lines breaks.
>
> One good example to see the problem happening: https://standard-reader.app/a/did:plc:2cxgdrgtsmrbqnjkwyplmp43/mr38ns2d8hf9qzw5mt
>
> By the way, Skyreader's Linkblogs are probably using markdown (quotes start with ">").

## What I found

Skyreader doesn't define its own lexicon for linkblogs — it reuses `pub.leaflet.content` (confirmed by fetching the linked document's record directly from the author's PDS). But instead of using Leaflet's structured blocks (a `blockquote` block, separate `text` blocks per paragraph), it writes raw markdown — blank-line-separated paragraphs and a leading `"> "` blockquote line — into the `plaintext` of a single `pub.leaflet.blocks.text` block.

`renderer-core`'s Leaflet path (`leafletToNode` in `build.ts`) mapped a `text` block's `plaintext` straight onto one `paragraph` node, verbatim, with no markdown parsing. The app's paragraph CSS has no `white-space: pre-wrap`, so the embedded `\n\n` collapsed under normal HTML whitespace rules — hence "one big clump of text" — and the literal `"> "` prefix rendered as plain text instead of a blockquote.

The reporter's guess about markdown was right, but the actual fix isn't routing this through the markdown-in-record pipeline (`markdown.ts`) — that pipeline is keyed off a registry of known markdown `$type`s, and `pub.leaflet.content` deliberately isn't in it, since real Leaflet documents use structured blocks, not markdown-in-plaintext. Instead, I added a small, format-scoped fixup: [`leaflet/plaintext-paragraphs.ts`](https://github.com/hipstersmoothie/standard-reader/blob/claude/feedback-3mruxyev4mcb3/packages/renderer-core/src/leaflet/plaintext-paragraphs.ts) splits a Leaflet `text`/`blockquote` block's `plaintext` on blank lines into separate `paragraph`/`blockquote` nodes, and converts a leading `"> "` line into a real `blockquote` node. Byte-offset facets (the two `didMention` facets in the example document) are reindexed onto whichever split segment they land in; a facet split across two segments is dropped rather than rendered against the wrong text.

Verified against the real record fetched from the author's PDS (`did:plc:2cxgdrgtsmrbqnjkwyplmp43`, rkey `mr38ns2d8hf9qzw5mt`): the leading quoted paragraph now renders as a `blockquote`, the following three paragraphs split apart correctly, and both `@mention` facets still resolve to the right substring after the split.

This is scoped tightly to the Leaflet `text`/`blockquote` block path — genuine Leaflet content (blocks whose `plaintext` never contains a blank line, which is true of every WYSIWYG-authored block) round-trips through unchanged, so this is a no-op for the common case.

### Why this skipped the `craft` design flow

This is filed as `bug`, and the symptom is visible on the page, but the fix is a content-parsing correction in the shared `renderer-core` package, not a new UI/visual design decision — no components, styling, spacing, or layout choices changed. It reuses the existing `Paragraph` and `Blockquote` components/styles exactly as they already render for genuine multi-paragraph/blockquote Leaflet documents elsewhere in the app.

## Testing

- Added unit tests in `packages/renderer-core/src/__tests__/build.test.ts` covering: blank-line paragraph splitting, a leading `"> "` line becoming a blockquote with facets correctly reindexed, and that single-paragraph plaintext is left untouched.
- Manually verified against the real production record (fetched from the PDS) with a standalone script — confirmed the blockquote/paragraph split and correct facet-to-mention mapping.
- `pnpm build && pnpm lint && pnpm typecheck && pnpm test` all pass.

---
_Generated by [Claude Code](https://claude.ai/code/session_01AcVZYLgbJozMLg4XwUAVdz)_


hipstersmoothie:
Status note on the recurring Railway per-service deploy-status failures on this preview environment (commit `2592d35`):

- **`recompute-cron`** — pre-existing crash/auto-restart cycle unrelated to this PR (diagnosed earlier: it ran clean for ~19h on the original unchanged commit before first crashing, and has continued a crash → auto-restart → recover loop since). This PR's diff (`packages/renderer-core`, `apps/standard-reader/TODO.md`) doesn't touch anything `recompute-cron` depends on.
- **`claudeslop`** / **`botlabeler`** — build failures unrelated to this PR; `services/claudeslop` and the botlabeler service have zero dependency on `packages/renderer-core`. `botlabeler`'s failure specifically correlates with a failed `Point Railway preview at Neon branch` CI step (preview DB provisioning), not this diff.
- **`ingest`** — crashed once (~2:33pm UTC) and, unlike `recompute-cron`, hasn't self-recovered after several hours. Also not caused by this diff (confirmed via `git diff` — this PR touches nothing under `services/` or the ingest path). Recent main-branch commits merged into this branch changed ingest's failure-handling behavior itself (e.g. "Make a wedged ingest channel fail instead of going quiet"), which may explain the new crash pattern in this preview env, but that's main's code, not this PR's.

The actual required checks — **Lint, format, typecheck, build** and **Lint & publish lexicons** — are green on this commit. I don't have the means to re-run the Railway per-service deploys myself; will keep watching but won't re-comment on further recurrences of these same known issues unless something changes.

---

PR #123, opened 2026-08-01
Give repo repair its own service, and a daily lap

hipstersmoothie:
## Why

A reader reported an article that was live on their PDS but missing from Reader: [`site.standard.document/3mrzba3giok2m`](https://standard-reader.app/a/did:plc:bpotnohnlgcj3fbmp7ugx4en/3mrzba3giok2m) on `@youronly.one`.

The record was fine. tap had the repo at a rev *past* the commit that created it and reported itself perfectly healthy — `state: "active"`, `error: ""`, `retries: 0`. There was no `ingest_dead_letter` row and no error log, because a dropped event and a quiet repo are indistinguishable downstream. Sampling 25 recently-active repos against their PDSes found the same gap on several more (crownnote.com was missing 186 documents).

The backstop for exactly this already exists — `repairRepoIfAdvanced` compares each repo's head rev to `last_seen_rev` and re-mirrors the drift. It just couldn't get there in time. Between the ingest worker's 30-minute timer (5 repos) and a batch of 50 on each hourly recompute, the round-robin was doing **~100 repos/hour against ~10,000 tracked — a four-day lap**. `youronly.one`'s `tracked_repos.updated_at` was 16 hours stale when the report came in.

## What changed

**Repair moves off the ingest worker into its own cron service.** Repairing a repo means enumerating it and rewriting every changed record, so running it beside the live tap channels put the backstop in direct competition with the stream it backstops — on a worker that already sits pinned at `inflight=12` (its cap) on nearly every heartbeat. Note the contrast with `recompute-cron`, which is only a *trigger*: it POSTs the ingest worker and the work still runs there. This one does its work in its own process, which is the whole point.

**The batch is derived from the fleet, not fixed.** A constant can't hold a lap length as the fleet grows, so: one lap is `RECONCILE_LAP_HOURS`, the cron fires every `RECONCILE_INTERVAL_HOURS`, and it takes that fraction of the fleet. At today's 10,359 repos the default daily lap is **432/hour**, taking the worst case for an unnoticed missing article from ~4 days to **~24 hours**.

I left `RECONCILE_CONCURRENCY` at 8 and only dropped the half of its comment that justified it by "the same worker is serving live ingest" — no longer true. The other half still holds (most repos cluster on a few PDS hosts), and 432/hour finishes in minutes at 8-deep.

### Two new logs, so next time we don't infer this from a hand-run diff

- **`ingest.repoStreamGap`** — records the repair had to insert *from scratch*, which is the only measurement of the live stream's loss rate we have. Deliberately narrower than `upsertedDocuments` (which also counts stale-CID updates — a missed *update* is less user-visible than a missed *create*). Counted by URI presence rather than CID, so rows soft-deleted by `dedupeRecords` don't read as a permanent phantom gap. Reported next to the repo's total, because the ratio is what separates stream loss from a first backfill.
- **`ingest.recordDropped`** — `handleRecord`'s two paths that return without doing anything while the caller acks the event as applied: a create/update with no record body, and an unmodeled collection. The latter logs once per collection per process — if tap's filter ever slips, that arrives at firehose volume.

## Verification

Smoke-run against prod with `RECONCILE_BATCH=12`: attempted 12, 8 enumerated, 141s. `repoStreamGap` fired on three real repos — **4/118, 8/4337, 36/12079** documents. Small ratios on large repos: dropped events, not missing backfills.

Also repaired the reported article and the other affected publishers by hand while diagnosing, so those are already live.

Typecheck, format, lint clean; 11 ingest tests pass.

## Deploy — needs a Railway step

1. New service in `standard-reader`, same repo/branch.
2. **Settings → Config-as-code → `railway.reconcile.json`.** Without this Railway silently deploys the web build (the known footgun).
3. Env: `DATABASE_URL`, `TAP_API_URL`, `TAP_ADMIN_PASSWORD`, `HONEYCOMB_API_KEY`. It needs tap's credentials because repairing a record can newly track its author's repo.

⚠️ **Until that service exists, nothing schedules repair at all** — the ingest-side timers are gone in this PR. Worth landing both together.

## Found along the way, not fixed here

Four repos fail repair with their PDS returning `400 InvalidRequest` on `listRecords` — records whose `tags` exceed the lexicon's `maxGraphemes: 128`, so the PDS won't serve the collection at all. The failure backoff handles it, but those repos can *never* be repaired: enumeration fails before it starts, so any event tap drops for them is permanent. Separate issue.

### DIFF
--- a/.agents/skills/railway-inspect/SKILL.md
+++ b/.agents/skills/railway-inspect/SKILL.md
@@ -18,9 +18,14 @@ alone (they describe intent; the CLI shows reality).
 
 ## Standard Reader Railway topology
 
-The `standard-reader` Railway project runs four long-lived services plus an
-hourly cron, sharing a single Neon Postgres read-model. GitHub auto-deploys on
+The `standard-reader` Railway project runs four long-lived services plus two
+hourly crons, sharing a single Neon Postgres read-model. GitHub auto-deploys on
 push to `main`. Every service uses the `RAILPACK` builder.
+
+Note the split between the two crons: `recompute-cron` is only a *trigger* — it
+POSTs the ingest worker and the work runs there. `reconcile-cron` does its work
+in its own process, on purpose, so repo repair never competes with the live tap
+stream it exists to backstop.
 
 | Service          | Start command                                      | Config file            | Healthcheck / notes                                                   |
 | ---------------- | -------------------------------------------------- | ---------------------- | --------------------------------------------------------------------- |
@@ -28,6 +33,7 @@ push to `main`. Every service uses the `RAILPACK` builder.
 | `tap`            | `ghcr.io/.../tap` Docker image on a `/data` volume | (Docker, no railpack)  | Long-running firehose consumer; admin API on `:2480` (private)         |
 | `ingest`         | `pnpm ingest:start` (= `tsx src/server/ingest/service.ts`) | `railway.ingest.json`  | Binds `[::]:3099`; consumes `tap.railway.internal:2480`               |
 | `recompute-cron` | `node scripts/recompute-cron.mjs`                 | `railway.cron.json`    | `0 * * * *`; POSTs the ingest worker's `/api/ingest/recompute`; `restartPolicyType: NEVER` |
+| `reconcile-cron` | `pnpm reconcile:repos`                            | `railway.reconcile.json` | `20 * * * *`; PDS repair round-robin, runs in its own process (does **not** call ingest); `restartPolicyType: NEVER` |
 
 **Runbook gotcha:** Railway auto-detects only the root `railway.json`. Every
 non-web service needs its **Config File Path** set explicitly (Dashboard →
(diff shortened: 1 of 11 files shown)


hipstersmoothie:
Heads-up on the CI failure here — it wasn't from this branch.

`pnpm lint` has been red on `main` since fd8a5662 ("speed up latest"), which committed `apps/standard-reader/scripts/tmp-drain.ts` carrying two oxlint violations (`unicorn/numeric-separators-style` and `unicorn/no-process-exit`). [That run failed](https://github.com/hipstersmoothie/standard-reader/actions/runs/30709791805) and every PR opened on top of it inherits the break. Fixed in 7b62df98 — numeric separator on the minutes divisor, plus the same `no-process-exit` disable every other script in that directory uses.

Worth a look separately: `tmp-drain.ts` describes itself as "One-off: drain the first-pass repair backlog", and it selects exactly the repos this PR's round-robin now visits continuously (`last_seen_rev is null`, non-gone, non-bridged). It looks like it rode along in an unrelated commit and is now subsumed. I left it in place rather than delete it — your call.

PR #140, opened 2026-08-09
Notify readers when a publication or author they pick publishes

hipstersmoothie:
Standard Reader had no notifications at all. The only way to learn that a source you care about had published was to open the app and look, or wait for the Monday digest. This adds a bell to publication and author pages: turn it on, and the next time that source publishes, your browser tells you.

**Notifications are deliberately independent of subscribing.** The bell doesn't change your feed and doesn't require a subscription.

## The process split is the design

The tap worker's *entire* involvement is one `INSERT … SELECT` that resolves to zero rows for anything that doesn't qualify — no timer, no `web-push` import, no outbound HTTP, no extra pool pressure. A new `push-cron` Railway service does the fan-out, encryption, and delivery every five minutes in its own short-lived process, the same shape as the weekly digest.

That's not fastidiousness: the worker's job is keeping up with the firehose, and a slow handler there has already knocked it over once by causing tap to redeliver.

## Exactly-once, and why it's a table

`document_push_queue` is both the outbox and the claim. Its primary key is what makes tap redelivery, dead-letter replay, and later `update` events for the same document all collapse to `DO NOTHING`.

It is deliberately **not** a column on `documents`. Document deletes are hard deletes (`db.delete(documents)`, not a tombstone), so a claim token living on that row would die with it — and a delete→recreate at the same rkey is a normal republish, which would re-notify everyone.

## Four guards, four different failure modes

| Guard | What it actually prevents |
| --- | --- |
| `payload.live` | A PDS reconcile replaying a repo notifying about its whole history |
| A date on the **record** | `documents.published_at` falls back to `now()` for an undated record, so an archive import would pass any freshness window |
| `indexed_at` | Set once on first insert, never updated — survives a tap cursor rewind or a fresh tap instance re-streaming known repos |
| Per-author hourly cap | The backstop for a repo that dumps its archive with fresh timestamps |

Note the enqueue is **not** gated on `action === "create"`. `upsertDocument` resolves external bodies (Leaflet, pckt, fetched formats); a transient failure there leaves `hasRenderableBody = false`, so a create-only gate would find nothing and the post would *never* notify, because the later `update` that fixes the body wouldn't retry.

## Service worker

Push listeners ride into the existing Workbox SW via `workbox.importScripts` rather than switching to `injectManifest`, which would mean hand-porting the tuned runtime-caching config and risking the offline story.

`push-sw.js` is **also** precached, deliberately. `importScripts` bakes a bare URL into `sw.js`, so editing the imported file changes zero bytes of `sw.js` and browsers keep running the old push handler indefinitely. Precaching puts its revision hash in the manifest, so the SW byte-diffs and the update propagates. Verified in the build output:

```
importScripts("/push-sw.js")
precache: push-sw.js -> 807f433486
```

It's a classic script with no imports and no top-level `await`, and it never throws — Workbox emits the SW as AMD, so a throw there would take down every caching route with it. And it shows a notification on *every* push, including an unparseable one: iOS revokes the subscription if a push arrives and nothing is displayed.

## iOS

The bell stays visible and pressable. iOS only exposes push to a Home-Screen-installed PWA, and there's no programmatic install prompt, so pressing it opens the Share → Add to Home Screen steps. That's the one "unsupported" case a reader can actually fix, and a disabled control wouldn't tell them so.

## Verification

- `pnpm typecheck`, `pnpm lint`, `pnpm format:check` — clean across the monorepo
- `pnpm test` — 767 passed, 6 skipped; 20 new tests covering the enqueue date guard, tag collapsing, name fallbacks, and payload truncation
- `pnpm build` — succeeds; SW wiring verified in the output as above
- SQL parameter binding checked read-only against real Postgres. This repo has been bitten by `sql` templates that typecheck and pass tests but bind wrong, so `make_interval(hours => $1)` was confirmed to produce the right offsets, the enqueue predicate to match real rows, and a zero-hour control to match none — proving the parameter is genuinely applied. Recipient resolution uses `inArray`/`notInArray` rather than raw `= any(...)` for the same reason.

## Before this can be merged

1. Create the `push-cron` Railway service with its **Config File Path set to `railway.push.json`** — Railway auto-detects only the root `railway.json` and will otherwise silently deploy the web build.
2. Generate a VAPID pair: `pnpm --filter standard-reader exec web-push generate-vapid-keys`
3. Set `VAPID_PUBLIC_KEY` on **web**; set all three VAPID vars plus `PUBLIC_URL` on **push-cron**. The ingest worker needs nothing.

With the keys unset the cron exits cleanly having done nothing, so a preview environment pointed at a shared database can't send real notifications carrying preview-domain links.

### DIFF
--- a/apps/standard-reader/.env.example
+++ b/apps/standard-reader/.env.example
@@ -298,6 +298,37 @@ CLAUDE_ROUTINE_TOKEN=""
 # DIGEST_SEND_DELAY_MS="1500"  # throttle between sends
 
 # ──────────────────────────────────────────────────────────────────────────────
+# Web push notifications (VAPID)
+# ──────────────────────────────────────────────────────────────────────────────
+# "Notify me when this publication/author posts". The tap worker queues a row
+# the instant a new document is indexed; the self-contained `push-cron` Railway
+# service drains and sends it via `pnpm push:send` every five minutes.
+#
+# Which service needs what:
+#   push-cron  — all three below, PLUS DATABASE_URL and PUBLIC_URL. This is the
+#                only process that ever holds the private key.
+#   web        — VAPID_PUBLIC_KEY only, so `getPushConfig` can hand the browser
+#                the application server key to subscribe with.
+#   ingest     — nothing. Its whole involvement is one INSERT.
+#
+# With the keys unset the cron exits cleanly having done nothing, so a PR
+# preview pointed at a shared database can't send real notifications carrying
+# preview-domain links.
+#
+# Generate a pair with: pnpm --filter standard-reader exec web-push generate-vapid-keys
+# VAPID_PUBLIC_KEY=""
+# VAPID_PRIVATE_KEY=""
+# Contact URL for the push services, `mailto:` or `https:`:
+# VAPID_SUBJECT="mailto:hello@standard-reader.app"
+#
+# Optional tuning:
+# PUSH_MAX_DOCUMENTS_PER_RUN="20"  # documents drained per invocation
+# PUSH_MAX_SENDS_PER_RUN="2000"    # ceiling on individual notifications per run
+# PUSH_SEND_CONCURRENCY="8"        # endpoints pushed to in parallel
+#
+# Local run: `pnpm push:send:dev` (loads this .env).
+
+# ──────────────────────────────────────────────────────────────────────────────
 # Topic naming (recompute sweep)
 # ──────────────────────────────────────────────────────────────────────────────
 # Server-only, needed by the `ingest` service. The sweep clusters the tag
--- a/apps/standard-reader/APP_VISION.md
+++ b/apps/standard-reader/APP_VISION.md
@@ -1147,6 +1147,18 @@ Standard Reader speaks the standard AT Proto label protocol, so readers can subs
 - **claudeslop** is our example labeler: a standalone service (`services/claudeslop/`) that
   consumes Jetstream, scores documents for AI-written prose, signs labels, and serves
   `queryLabels` + `subscribeLabels` — a minimal reference implementation of the labeler API.
+- **Web push notifications** are the one piece of personal state that is **app-owned, not
+  repo-owned** — a deliberate exception to the rule at the top of this section. A reader turns on
+  a bell from a publication or author page and gets a notification when that source publishes.
+  Three tables carry it (`src/db/schema/push.ts`), none of them mirroring a record:
+  `push_devices` holds a browser's push endpoint and encryption keys, which are per-browser
+  secrets — anyone holding them can push to that browser — and so must never land in a public
+  repo. `push_topics` (the per-source opt-in) _could_ have been a record like `sidebarPref`, but a
+  new record collection means an expanded OAuth write scope, which would put every existing
+  session through the "reconnect your account" flow before the bell worked at all; if the opt-in
+  ever needs to be portable, a record + mirror can be layered on the same table.
+  `document_push_queue` is the send outbox. **Notifications are independent of subscribing** —
+  the bell doesn't change your feed and doesn't require a subscription.
 
 ---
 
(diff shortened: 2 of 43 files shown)


hipstersmoothie:
## Railway is configured — ready to verify

**Applied to production:**

| What | State |
| --- | --- |
| Migration `0033` (3 new tables) | ✅ applied to prod |
| `web` → `VAPID_PUBLIC_KEY` | ✅ set (public key only) |
| `push-cron` service | ✅ created, deploying from `feat/web-push-notifications` |
| `push-cron` **Config File Path** | ✅ `railway.push.json` |
| `push-cron` deploy | ✅ `SUCCESS` — `start=pnpm push:send`, `cron=*/5 * * * *`, `restart=NEVER` |
| `ingest` | untouched, 0 VAPID vars — as designed |

`push-cron` reads `DATABASE_URL` / `PUBLIC_URL` / `VAPID_PUBLIC_KEY` as Railway references to `web` rather than copies, so there's one place to rotate them. Only `VAPID_PRIVATE_KEY` is stored directly, and only there.

Worth recording: `railway add` triggers a deploy **immediately**, before you can set the Config File Path — and the CLI can't set that field at all. The first two deploys came up as `start=pnpm start` (the web build running as a long-lived service). Both were removed and redeployed once `railwayConfigFile` was set via `serviceInstanceUpdate` on the GraphQL API. Anyone adding a service to this monorepo will hit the same ordering trap.

**Verified against the production database** (probe rows inserted and cleaned up):

- All three tables created with the expected shape
- The enqueue `INSERT … SELECT` returns the row
- A second identical insert returns **0 rows** — exactly-once holds
- The `FOR UPDATE SKIP LOCKED` claim works and increments `attempts`
- An immediate re-claim returns **0** — the reaper correctly does *not* steal a fresh claim
- `make_interval(hours => $1)` binds correctly (24h/10min offsets exact), the enqueue predicate matches real documents, and a zero-hour control matches none — proving the parameter is genuinely applied, not silently ignored

## One gap before this is end-to-end

The **enqueue lives in the ingest worker**, which is still running `main`. So nothing will land in `document_push_queue` from live traffic until this merges. The send half is exercisable now (a queued probe row is waiting for the next cron tick), but the firehose → queue half only starts working post-merge.

I did **not** point `ingest` at this branch — it's the critical firehose consumer and that's not a change to make casually. Say the word if you'd rather I did, for a true end-to-end test before merge.

## To try the UI

The `web` redeploy carries the VAPID public key but *not* this branch's code, so the bell isn't on the site yet either. For a local run: `pnpm build && pnpm preview` (the service worker is disabled in `vite dev`), then the bell appears on any publication or author page.

hipstersmoothie:
## Rebased — and two things worth flagging

**The branch was based on `perf/incremental-topic-counts`, not `main`.** It carried two unrelated commits and the PR was `CONFLICTING`, which is also why GitHub Actions never ran on the first push. Rebased onto `main`; it's now one commit and `MERGEABLE`.

### The migration collided, and prod's ledger had to be repaired

`main` had already landed its own `0033` (`exclude_web_bridge`), so mine was renumbered to **`0034_nostalgic_hairball`**. That's more than a rename: the regenerated file gets a new hash *and* a new `when` timestamp, and I'd already applied the old `0033` to prod by hand. Prod's `drizzle.__drizzle_migrations` was left holding a row for a file that no longer exists in the repo.

Left alone, **the next production deploy would have failed** — `drizzle-kit migrate` would have seen `0034` as unapplied and re-run `CREATE TABLE` against tables that were already there. `drizzle-kit check` passes either way; it validates the journal against the repo and knows nothing about prod's ledger.

Repaired: confirmed the three tables held no user data (0 devices, 0 topics, 1 of my own probe rows), dropped them, deleted the stranded ledger row, and re-ran `pnpm db:migrate` so `0034` applied exactly as it will on deploy. Prod's ledger max now matches the journal.

One cosmetic side effect: a cron tick landed inside that drop/re-apply window and shows as a `CRASHED` deploy with a `relation does not exist` error. That's the repair window, not a code fault — the runs before and after are clean.

### CI caught a real miss

`Lint, format, typecheck, build` failed on stale translation catalogs — I'd added Lingui strings without running `pnpm i18n:extract`. Fixed, and while regenerating I noticed two of my strings differed only by a straight vs curly apostrophe, producing two catalog entries for one sentence. Normalized to the curly form the rest of the app uses, so they dedupe to a single entry referenced from both files. Re-running extraction is now idempotent.

### Still green after the rebase

`typecheck` · `lint` · `format:check` clean; 20 push tests pass. The `push-cron` service redeployed from the rebased commit and its latest run is clean:

```
[push] done in 510ms: {"claimed":0,"delivered":0,...}
```

Earlier, before the repair, a run picked up a queued probe document and settled it correctly — `state=sent, attempts=1, last_error=null` — so the full claim → resolve → settle loop is confirmed working in production, not just in tests.

`recompute-cron` also passes now, which confirms its earlier failure was a preview-environment timeout talking to `ingest`, unrelated to this change.

hipstersmoothie:
## Can the PR preview deliver real pushes? Yes — with one manual step

**All 12 CI checks pass.**

### What works in the preview automatically

| Piece | State |
| --- | --- |
| Preview web (`web-standard-reader-pr-140.up.railway.app`) | runs this branch; `VAPID_PUBLIC_KEY` now set + redeployed |
| `/push-sw.js` | served ✅ |
| Preview Neon branch | already has all three tables (`0034` applied) **and 3M real documents** — it forked from prod after the migration |
| Preview `ingest` | runs this branch against the preview DB, so live firehose documents **will** enqueue there |

So the bell, permission prompt, device registration, topic opt-in, and enqueue all work on the preview as-is.

### The one gap: nothing drains the queue there

`push-cron` has **no service instance** in the PR-140 environment. Railway forked that environment before the service existed, and setting variables on it creates a variable record but not a deployable instance — `serviceInstanceDeployV2` returns `Service Instance not found`. Any *future* PR env will include it, since the service now exists in the project and `railway.push.json` is on the branch.

**Workaround — this delivers genuinely real pushes:**

1. Open the preview, sign in, hit the bell on a publication or author (grant the permission prompt).
2. Run the sender locally against the preview database. It's the identical code path the cron runs:

```bash
DATABASE_URL='<preview branch URL>' \
DB_DRIVER=pg \
PUBLIC_URL='https://web-standard-reader-pr-140.up.railway.app' \
VAPID_PUBLIC_KEY='<public>' VAPID_PRIVATE_KEY='<private>' \
VAPID_SUBJECT='mailto:hello@standard-reader.app' \
pnpm --filter standard-reader exec tsx scripts/send-push.ts
```

Your browser gets a real notification, and clicking it opens the article on the preview.

The only thing that step skips is the five-minute scheduler — and that half is already proven in production: a queued probe document was claimed and settled `state=sent, attempts=1, last_error=null`, and subsequent runs log clean (`[push] done in 510ms`).

Alternatively, closing and reopening the PR rebuilds the preview environment and would clone `push-cron` into it, giving you the scheduled path too.

### Timing note

The queue only fills when a source you've belled **actually publishes**, which could be a while. Happy to queue a row for a recent document from whichever publication you turn the bell on for, so there's something to deliver immediately.

PR #141, opened 2026-08-09
Notify readers when a publication or author they pick publishes

hipstersmoothie:
Standard Reader had no notifications at all. The only way to learn that a source you care about had published was to open the app and look, or wait for the Monday digest. This adds a bell to publication and author pages: turn it on, and the next time that source publishes, your browser tells you.

**Notifications are deliberately independent of subscribing.** The bell doesn't change your feed and doesn't require a subscription.

> Supersedes #140, closed so this PR gets a fresh preview environment that includes the new `push-cron` service.

## The process split is the design

The tap worker's *entire* involvement is one `INSERT … SELECT` that resolves to zero rows for anything that doesn't qualify — no timer, no `web-push` import, no outbound HTTP, no extra pool pressure. A `push-cron` Railway service does the fan-out, encryption, and delivery every five minutes in its own short-lived process, the same shape as the weekly digest.

That's not fastidiousness: the worker's job is keeping up with the firehose, and a slow handler there has already knocked it over once by causing tap to redeliver.

## Exactly-once, and why it's a table

`document_push_queue` is both the outbox and the claim. Its primary key is what makes tap redelivery, dead-letter replay, and later `update` events for the same document all collapse to `DO NOTHING`.

It is deliberately **not** a column on `documents`. Document deletes are hard deletes (`db.delete(documents)`, not a tombstone), so a claim token on that row would die with it — and a delete→recreate at the same rkey is a normal republish, which would re-notify everyone.

## Four guards, four different failure modes

| Guard | What it actually prevents |
| --- | --- |
| `payload.live` | A PDS reconcile replaying a repo notifying about its whole history |
| A date on the **record** | `documents.published_at` falls back to `now()` for an undated record, so an archive import would pass any freshness window |
| `indexed_at` | Set once on first insert, never updated — survives a tap cursor rewind or a fresh tap instance re-streaming known repos |
| Per-author hourly cap | The backstop for a repo that dumps its archive with fresh timestamps |

The enqueue is **not** gated on `action === "create"`. `upsertDocument` resolves external bodies (Leaflet, pckt, fetched formats); a transient failure leaves `hasRenderableBody = false`, so a create-only gate would find nothing and the post would *never* notify, because the later `update` that fixes the body wouldn't retry.

## Service worker

Push listeners ride into the existing Workbox SW via `workbox.importScripts` rather than switching to `injectManifest`, which would mean hand-porting the tuned runtime-caching config.

`push-sw.js` is **also** precached, deliberately. `importScripts` bakes a bare URL into `sw.js`, so editing the imported file changes zero bytes of `sw.js` and browsers keep the old push handler indefinitely. Precaching puts its revision hash in the manifest so the SW byte-diffs. Verified in the build output:

```
importScripts("/push-sw.js")
precache: push-sw.js -> 807f433486
```

It's a classic script with no imports and no top-level `await`, and it never throws — Workbox emits the SW as AMD, so a throw there would take down every caching route. It shows a notification on *every* push, including an unparseable one: iOS revokes the subscription if a push arrives and nothing is displayed.

## Two bugs found by testing on the previous preview

**The bell always failed** with "we couldn't set up notifications on this device". `getPushConfig` gated on a `configured` getter requiring *both* VAPID keys — but the web service holds only the public key by design. It always returned a null key, which the client read as an unsupported browser. Now split into `canRegister` (public key — what a browser needs to subscribe) and `canSend` (both — what signing needs), with tests pinning each service's real state. The client also collapsed "server has no key" into the same toast as a genuine exception, which made a deployment mistake look like a browser problem; that's now a distinct state, and both failure paths log the real cause.

**The bell was wedged inside the Subscribe `ButtonGroup`**, which joins its children into one control — so it landed between Subscribe and Subscribe's own dropdown chevron, breaking that pairing, and took the app's neutral variant while the rest of the group took the publication's theme accent (visible on "What The Function!?"). It now sits beside the group and takes the same variant.

## iOS

The bell stays visible and pressable. iOS only exposes push to a Home-Screen-installed PWA and offers no programmatic install prompt, so pressing it opens the Share → Add to Home Screen steps — the one "unsupported" case a reader can actually fix.

## Verification

- `pnpm typecheck` · `lint` · `format:check` clean across the monorepo
- 24 push tests; full suite 767 passing
- `pnpm build` succeeds; SW wiring verified in the output
- **Confirmed in production, not just tests:** a queued document was claimed and settled `state=sent, attempts=1, last_error=null`; subsequent cron runs log clean (`[push] done in 510ms`)
- SQL parameter binding checked read-only against real Postgres — this repo has been bitten by `sql` templates that typecheck and pass tests but bind wrong. `make_interval(hours => $1)` produces the right offsets, the enqueue predicate matches real documents, and a zero-hour control matches none. Recipient resolution uses `inArray`/`notInArray` rather than raw `= any(...)` for the same reason.
- Exactly-once verified against prod: a second identical insert returns 0 rows, and an immediate re-claim returns 0 (the reaper doesn't steal a fresh claim)

## Deployment state

Production is already wired: migration applied, `VAPID_PUBLIC_KEY` on `web`, and a `push-cron` service with its **Config File Path set to `railway.push.json`** (Railway auto-detects only the root `railway.json`; the CLI can't set this field, so it goes through `serviceInstanceUpdate` on the GraphQL API). After merge, `push-cron` needs pointing from this branch to `main`.

### DIFF
--- a/apps/standard-reader/.env.example
+++ b/apps/standard-reader/.env.example
@@ -298,6 +298,37 @@ CLAUDE_ROUTINE_TOKEN=""
 # DIGEST_SEND_DELAY_MS="1500"  # throttle between sends
 
 # ──────────────────────────────────────────────────────────────────────────────
+# Web push notifications (VAPID)
+# ──────────────────────────────────────────────────────────────────────────────
+# "Notify me when this publication/author posts". The tap worker queues a row
+# the instant a new document is indexed; the self-contained `push-cron` Railway
+# service drains and sends it via `pnpm push:send` every five minutes.
+#
+# Which service needs what:
+#   push-cron  — all three below, PLUS DATABASE_URL and PUBLIC_URL. This is the
+#                only process that ever holds the private key.
+#   web        — VAPID_PUBLIC_KEY only, so `getPushConfig` can hand the browser
+#                the application server key to subscribe with.
+#   ingest     — nothing. Its whole involvement is one INSERT.
+#
+# With the keys unset the cron exits cleanly having done nothing, so a PR
+# preview pointed at a shared database can't send real notifications carrying
+# preview-domain links.
+#
+# Generate a pair with: pnpm --filter standard-reader exec web-push generate-vapid-keys
+# VAPID_PUBLIC_KEY=""
+# VAPID_PRIVATE_KEY=""
+# Contact URL for the push services, `mailto:` or `https:`:
+# VAPID_SUBJECT="mailto:hello@standard-reader.app"
+#
+# Optional tuning:
+# PUSH_MAX_DOCUMENTS_PER_RUN="20"  # documents drained per invocation
+# PUSH_MAX_SENDS_PER_RUN="2000"    # ceiling on individual notifications per run
+# PUSH_SEND_CONCURRENCY="8"        # endpoints pushed to in parallel
+#
+# Local run: `pnpm push:send:dev` (loads this .env).
+
+# ──────────────────────────────────────────────────────────────────────────────
 # Topic naming (recompute sweep)
 # ──────────────────────────────────────────────────────────────────────────────
 # Server-only, needed by the `ingest` service. The sweep clusters the tag
--- a/apps/standard-reader/APP_VISION.md
+++ b/apps/standard-reader/APP_VISION.md
@@ -1153,6 +1153,18 @@ Standard Reader speaks the standard AT Proto label protocol, so readers can subs
 - **claudeslop** is our example labeler: a standalone service (`services/claudeslop/`) that
   consumes Jetstream, scores documents for AI-written prose, signs labels, and serves
   `queryLabels` + `subscribeLabels` — a minimal reference implementation of the labeler API.
+- **Web push notifications** are the one piece of personal state that is **app-owned, not
+  repo-owned** — a deliberate exception to the rule at the top of this section. A reader turns on
+  a bell from a publication or author page and gets a notification when that source publishes.
+  Three tables carry it (`src/db/schema/push.ts`), none of them mirroring a record:
+  `push_devices` holds a browser's push endpoint and encryption keys, which are per-browser
+  secrets — anyone holding them can push to that browser — and so must never land in a public
+  repo. `push_topics` (the per-source opt-in) _could_ have been a record like `sidebarPref`, but a
+  new record collection means an expanded OAuth write scope, which would put every existing
+  session through the "reconnect your account" flow before the bell worked at all; if the opt-in
+  ever needs to be portable, a record + mirror can be layered on the same table.
+  `document_push_queue` is the send outbox. **Notifications are independent of subscribing** —
+  the bell doesn't change your feed and doesn't require a subscription.
 
 ---
 
@@ -1289,6 +1301,32 @@ hand-tuned lists:
   publications, so the filtered Latest "All" badge gets its own precomputed `network_stats` scalar
   (`network_document_count_no_web_bridge`) rather than a live count, and the tag feed switches to a
   tag-first query shape (`selectTagArticleUris`) that the survival rate can't blow up.
+
+### Web push delivery
+
+Split across two processes, and the split is the whole design:
+
+- **The tap worker only enqueues.** `handleRecord`'s document branch runs one
+  `INSERT … SELECT` into `document_push_queue` (`src/server/push/enqueue.ts`) that inserts
+  nothing for anything that doesn't qualify. No timer, no `web-push` import, no outbound HTTP —
+  the worker's job is keeping up with the firehose, and it has been knocked over before by a
+  slow handler causing tap to redeliver.
+- **A `push-cron` Railway service sends** (`railway.push.json` → `pnpm push:send` →
+  `src/server/push/run.ts`), every five minutes, in its own short-lived process — the same shape
+  as the weekly digest, for the same reason.
+
+The queue table is also the **exactly-once claim**: its primary key is what makes tap
+redelivery, dead-letter replay, and later `update` events for one document all collapse to
+`DO NOTHING`. It is deliberately _not_ a column on `documents`, because document deletes are
+hard deletes — a delete→recreate at the same rkey is a normal republish, and a claim living on
+that row would die with it and re-notify everyone.
+
+Four guards decide whether a document is genuinely new, and each covers a different failure:
+`live` (so a PDS reconcile can't replay a repo's whole history), a date on the **record itself**
+(`documents.published_at` falls back to `now()` for an undated record, so an archive import
+would otherwise pass any freshness window), `indexed_at` (set once on first insert and never
+updated — this is what survives a tap cursor rewind or a fresh tap instance re-streaming known
+repos), and a per-author hourly cap in the sender as the backstop.
 
 ### Topic derivation
 
--- a/apps/standard-reader/TODO.md
+++ b/apps/standard-reader/TODO.md
@@ -32,8 +32,8 @@ Check items off as they land.
   - **Runbook gotcha:** Railway auto-detects only the root `railway.json`, so every non-web service
     in this monorepo needs its **Config File Path** set explicitly (Dashboard → service → Settings →
     Config-as-code, or `serviceInstanceUpdate{ railwayConfigFile }` via the GraphQL API) to
-    `railway.ingest.json` / `railway.cron.json` / `railway.reconcile.json`; otherwise it silently
-    falls back to the web build.
+    `railway.ingest.json` / `railway.cron.json` / `railway.reconcile.json` / `railway.push.json`;
+    otherwise it silently falls back to the web build.
     Shared `INGEST_WEBHOOK_SECRET` = `TAP_ADMIN_PASSWORD`; `PUBLIC_URL=https://standard-reader.app`;
     `ATPROTO_PRIVATE_KEY_JWK` is the ES256 private JWK. Prod DB was reset to a clean schema (drop
     `public` + `drizzle`, then `pnpm db:migrate`) before first backfill.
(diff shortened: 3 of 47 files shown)


hipstersmoothie:
## Preview is fully wired — nothing left for you but the bell

Closing #140 did the trick: the fresh environment cloned `push-cron` **with its Config File Path intact**, which is what was missing before.

| Piece | State |
| --- | --- |
| Preview web | `https://web-standard-reader-pr-141.up.railway.app` — deployed, `VAPID_PUBLIC_KEY` present from first boot |
| Preview `push-cron` | deployed: `start=pnpm push:send`, `cron=*/5 * * * *` ✅ |
| Preview `ingest` | runs this branch → live firehose documents enqueue into the preview DB |
| Preview Neon branch | all three tables + 3.0M real documents |
| `PUBLIC_URL` on preview cron | repointed to the pr-141 domain, so notification clicks open the preview and not prod |

**To test:** open the preview, sign in, hit the bell on a publication or author, accept the browser prompt. When that source publishes, you get a real notification within five minutes — no manual command needed this time.

The bell fixes from testing #140 are in: registration no longer fails on a correctly configured deployment, and the bell now sits beside the Subscribe split button (not wedged between it and its own chevron) taking the publication's theme variant.

### CI

11 of 12 pass. `recompute-cron` fails with a `fetch` timeout POSTing to `ingest.railway.internal` — that's a fresh-preview-environment race where the cron fires before `ingest` is accepting connections. Same failure appeared on #140 before the rebase and cleared on its own; it touches nothing in this change (`recompute-cron` only triggers the ingest worker over private networking).

### One caveat on timing

The queue only fills when a source you've belled **actually publishes**. If you'd rather not wait, tell me which publication you turned the bell on for and I'll queue one of its recent documents so the next cron tick delivers immediately.

hipstersmoothie:
## I disproved my own hypothesis — and the fix is diagnosis, not a refactor

`no service worker ever became active` on an installed iPhone PWA. My suspicion was that
`importScripts("/push-sw.js")` was killing the worker install: Workbox emits it inside the
module factory, so on a cold launch it's a live network fetch, and an iOS Home Screen app always
installs fresh (standalone uses a storage partition separate from Safari) — often over cellular.

**That's wrong, and I tested it rather than shipping it.** Evaluating the built `sw.js` in a VM
with a throwing `importScripts`:

```
SYNC script evaluation threw: NO — completed cleanly
LATER: unhandled rejection -> NetworkError fetching /push-sw.js
```

Script evaluation completes; the failure surfaces only as an unhandled rejection inside the AMD
factory. So the worker **still installs and activates** — it would be left *inert*, not absent.
That cannot produce the reported symptom, so **I did not port the worker to `injectManifest`.**
That refactor would have meant hand-porting the tuned runtime-caching config on a guess, and the
guess was wrong.

*(It does reveal a separate, quieter bug: a transient `/push-sw.js` fetch failure leaves a
permanently inert worker — no precache routes, no push handlers — until the next SW update.
Real, but not this. Filed mentally as follow-up, not fixed here on another guess.)*

## What actually shipped

**Registration failures were being thrown away.** vite-plugin-pwa ends registration in
`.catch(e => onRegisterError?.(e))` and `registerServiceWorker` never passed that callback. A
worker that failed to install produced *no signal at all* — the only symptom was a push timeout
long afterwards. Both callbacks are now wired, logged, and captured. Not push-specific: a failed
install takes every caching route down with it, so offline was silently broken too.

**The push flow assumed someone else had registered a worker.** It waited on one that, if
`<ReloadPrompt />`'s registration had failed, was never going to exist. It now registers the
worker itself when none is present, and waits on the registration's own state machine instead of
racing a blind timer — resolving the moment it activates, reporting immediately when a worker
goes `redundant` (a failed install) rather than burning the timeout, and allowing 45s instead of
10s for a cold install on cellular. That last one alone may be the whole story: 10s is not much
for an iPhone fetching `sw.js` + the Workbox chunk + `push-sw.js` on first launch.

**Settings → Notifications → Troubleshooting** now shows capability, standalone, permission,
service-worker state, subscription, and the last registration error — the literal strings the
code branches on. You're debugging on a phone, where devtools isn't a usable loop; a screenshot
of that panel answers the question directly.

## To test

Open the preview from your Home Screen app, go to **Settings → Notifications → Troubleshooting**
first and screenshot it, then tap a bell. If it fails again the panel says why, and `Last error`
will carry the registration failure that was invisible before.

Checks: `typecheck`, `lint`, `format:check` clean; 24 push tests pass; `offline.html` still
precached.

hipstersmoothie:
## The real bug: `/sw.js` has been shipping truncated — in production, since before this branch

Your Troubleshooting panel reported `SyntaxError: Unexpected end of script`, `service worker: none`, `subscription: no`. That's not a push bug at all.

| | bytes | parses? | ends with |
|---|---|---|---|
| Good local build | 2351 | ✅ | `…{statuses:[0,200]})]}),"GET")});` |
| **Preview** | 1774 | ❌ | `…maxEntries:300,maxAgeSeconds:2` |
| **Production** | 1745 | ❌ | `…ugin({statuses:[0,200]})]}),"G` |

Both served files are strict prefixes of a valid build, cut mid-token. `content-length` and the etag match the truncated size, so the artifact is broken **at rest**, not in transit. Production's copy contains **zero** references to `push-sw`, dating the breakage well before web push.

So no service worker has been registering at all. **Offline support and the "new version available" reload prompt have been silently dead**; push was just the first feature to notice — and the first with a diagnostics panel to say so.

I also have to correct myself: an earlier "parses OK" check of mine was wrong. I'd chained `&& echo "parses OK"` off `head`, not off `node --check`, so it reported success unconditionally. Checking the exit code properly is what exposed this.

### Mechanism

Two tools write `.output/public`: `vite-plugin-pwa` via its `outDir`, and Nitro, which owns that directory as its static root. The plugin's hooks also run **once per build pass** — I instrumented it, and it's three times:

```
precache  0 entries   → writes .output/public/sw.js
precache  0 entries   → writes .output/public/sw.js
precache  8 entries   → writes .output/public/sw.js
```

So the file is rewritten repeatedly while another tool assembles the same directory — the shape of a partial write, and consistent with two deploys truncating at *different* offsets while local builds are fine.

### What I shipped

`pnpm build` now fails if the emitted worker is unusable: it must parse, precache something, and include the offline fallback. Verified both ways — passes a good build, and rejects a byte-truncated one reproducing the exact production failure:

```
[verify-sw] service worker is not valid JavaScript (Unexpected end of input).
  1772 bytes, ends with: "…maxEntries:300,maxAgeSeconds:2"
```

**This deliberately can fail deploys.** That's the point: a loud failure beats silently shipping a dead service worker, and this build will now say definitively whether the truncation happens during the build (verifier fires) or after it (verifier passes but the served file is still broken — which would isolate it to deploy/serving). I chose to measure rather than guess at the interleaving, having already had one service-worker hypothesis disproved today.

### Also fixed

My commit had accidentally included `src/lib/guide/navigation.ts` from the concurrent comics-guide branch. That's stripped out — and it turned out to be the cause of the i18n CI failure too, since its 9 `msg` labels weren't in the catalogs. With it removed, extraction produces no diff. I'm now working in an isolated worktree so the comics session is untouched.

hipstersmoothie:
## Root cause found, and proven end to end on a real server

The file on disk was never the problem. **Nitro bakes a static-asset manifest into `.output/server/index.mjs`** at build time:

```
"/sw.js": { "size": 1774, "etag": "\"6ee-…\"", "path": "../public/sw.js" }
```

…while the finished worker is **2351 bytes**. At request time Nitro reads the *current* file but caps the response at the *recorded* size, so clients get a prefix cut mid-token.

That accounts for every observation: a deterministic, content-dependent cut; `content-length` and etag agreeing with the truncated size; a recorded etag that hashes to neither the full file nor its prefix (it belongs to an earlier, genuinely different file); and the part that proved the build wasn't at fault — **a Railway build that passed the artifact check I added an hour earlier while still serving garbage**.

Where the stale size comes from: `vite-plugin-pwa` runs its hooks once per build pass. Instrumenting a build shows three, emitting `0 entries`, `0 entries`, then the real `8 entries`. Nitro snapshots the directory partway through and records the zero-entry worker's size; the plugin then overwrites the file. `2351 − 1774 = 575` — the eight precache entries.

### Proof, not inference

Booted the actual built server and did a controlled A/B — same build, one variable changed:

| manifest `size` | served | `node --check` |
|---|---|---|
| `1774` (stale) | 1774 B, ends `…maxEntries:300,maxAgeSeconds:2` | **exit 1** — byte-for-byte the production failure |
| `2351` (repaired) | 2351 B, **byte-identical to the file on disk** | **exit 0** |

Two things worth recording from getting there:

- The first local attempt 404'd on `/sw.js`. That was a red herring — **every** root-level `.js` 404'd, including `push-sw.js` and `workbox-*.js`, which the repair never touches. The missing variable was `NODE_ENV=production`, which Railway sets and I hadn't. With it set, everything serves.
- I twice wrote `node --check f | tail; echo $?` and read the exit code of `tail` rather than of `node --check`. That is what made a truncated file look like it "parses OK" earlier in this PR. Checked directly now.

### The fix

`pnpm build` re-syncs any manifest entry whose recorded size disagrees with the file on disk, recomputing the etag so caches still invalidate, then asserts Nitro would serve the whole worker. Generalized across all public assets rather than special-casing `sw.js` — nothing about the race is specific to it — and it still fails the build outright if the worker can't parse or precaches nothing.

Also fixed a bug in the check I added an hour ago: it compared a recorded byte size against a JavaScript string length, which disagree because the worker contains multi-byte characters.

hipstersmoothie:
## Confirmed on the deployed preview — the service worker is intact for the first time

```
preview /sw.js  HTTP 200  bytes=2351
node --check    exit=0
ends with       …ponsePlugin({statuses:[0,200]})]}),"GET")});
```

Previously 1774 bytes and unparseable. It now serves the complete, valid worker, ending exactly where a good build ends.

This is the first time `/sw.js` has been servable — so registration should now succeed, which means the bell should work, and **offline support and the "new version available" prompt come back** after however long they've been dead.

### Please retest on the installed iPhone app

Worth noting: your Home Screen app has a failed registration cached. If the bell still misbehaves, remove the app from the Home Screen and re-add it — that clears the old registration state. Settings → Notifications → **Troubleshooting** should now read `service worker: active` rather than `none`, with no `Last error`.

### Production is still broken until this merges

`https://standard-reader.app/sw.js` is still serving 1745 truncated bytes. The repair lands there with this PR.

PR #149, opened 2026-08-10
Honour Bluesky blocks and block lists

hipstersmoothie:
A reader's blocks now follow them into Standard Reader, in both directions, with no setup and no import step. Same idea as the labeler port — but the mechanism is deliberately the opposite, for a reason worth reading before the diff.

## A block is a record, not a setting

`app.bsky.graph.block` already lives in the blocker's repo, and every AT Protocol app that honours a block reads that same record. So there is no `app.standard-reader.block` and there never should be: we mirror Bluesky's lexicon, and blocking from here writes the record the rest of the network already reads. Block someone in Standard Reader and they're blocked on Bluesky too, and wherever the reader goes next.

That's the opposite call from labeler subscriptions, and only because Bluesky keeps *those* in a preferences blob with no per-key write — which is why they had to be ported into records of our own. Nothing about blocks needs that.

One useful consequence: `com.atproto.repo.listRecords` is public, so **honouring** a reader's existing blocks costs no OAuth scope at all. Enforced from their first page load, whatever they granted. Only *writing* asks for consent.

## Four ways to be blocked, all equal in effect

| Direction | Record | Held by |
| --- | --- | --- |
| `blocking` | `app.bsky.graph.block` | viewer |
| `blocked-by` | `app.bsky.graph.block` | them |
| `list-blocking` | `listblock` + `listitem` | viewer |
| `list-blocked-by` | `listblock` + `listitem` | them |

All four hide content. That symmetry is what makes a block feel like a block rather than a mute — and `list-blocked-by` is Bluesky's `blockedByList`, the reason a reader can be blocked by someone who never named them.

## Sync: per reader, not the firehose

Blocks can't ride the tap. It tracks the repos we index; a block against one of our readers can be written by any repo on the network. So `server/blocks/sync.server.ts` pulls per reader on a 15-minute TTL, from four sources: their own records from their PDS, incoming blocks via Constellation backlinks, mod-list membership via the public AppView, and the listitem→listblock chain for `blockedByList`. Scheduled from the OAuth callback and the signed-in shell read; forced inline right after a block/unblock write so the change lands on the very next page.

**Outgoing replaces; incoming replaces only on a complete sweep.** The reader's own records are authoritative, so a sweep can soft-delete what's gone. Blocks *against* them come from a best-effort index, so their delete is gated on the sweep having read every page — a truncated or failed sweep adds but never removes, because it must not read as "nobody blocks you any more".

It does have to remove *eventually*, though, and the first cut of this got that wrong: incoming rows were additive-only, and the blocker is almost never a reader here, so nothing else would ever retire the row. An unblock made on Bluesky would have hidden that account from the reader permanently, with no path to clear it.

**A sweep never happens on a request anyone waits for.** A full sweep is dozens to hundreds of round trips. Blocking a list awaits only the two bounded reads the reader is actually waiting on — their own `listblock` records, and that list's membership — and schedules the rest; Settings → Refresh starts a sweep and says so rather than holding the request open. Sweeps are deduped per reader process-wide, so a doubled press joins the run already in flight.

## Enforcement

SQL `NOT EXISTS` (`notBlockedByViewer`) wherever a page paginates — home, latest, trending, tag, search, discover directory, publication archives, list feeds, XRPC feeds — so a block never shortens a page or strands results behind an offset. JS post-filter on rails, profile tabs, friend surfaces, and third-party discussion (Bluesky / Margin / Semble / pckt / Leaflet), where a shorter list is fine.

Every helper is bounded by the DIDs a page is about to render, never by the reader's block set. A reader on a 40k-member blocklist has to stay an index probe, which is what `block_list_items (list_uri, subject_did)` is for.

Both paths check the **author and the publication owner** — a guest post in a blocked owner's publication carries that owner in its byline. (`publicationOwnerDid` existed and was unit-tested from the start, but no production card actually carried the field, so that half of the filter was dead until `ArticleCard` grew it. In SQL the owner comes from the publication URI's authority, so it needs no join and works inside the lateral.)

### The SQL predicate is opt-in per reader

`notBlockedByViewer` is three correlated `NOT EXISTS` per candidate row, and on the follow feed it lands inside `baseWhere` — which is interpolated into the per-source laterals whose whole job is an indexed `ORDER BY published_at DESC LIMIT k` scan. That is precisely the shape the strategy note in `queries.ts` warns about, and precisely what caused the 1651ms corpus-walk regression that note exists to document.

So the only supported way to reach it is `blockFilterDid`, which answers "does this reader block anybody" from an in-process cache (60s TTL, invalidated on every write) and returns `undefined` for the overwhelming majority who don't — keeping the predicate out of the query entirely rather than paying for it on every signed-in feed load. `queries.explain.test.ts` (`pnpm perf:explain`) gained two cases asserting the *filtered* plan still rides the per-source composite indexes.

Detail routes (`getArticle`, `getPublicationProfile`, `getAuthorProfile`, the per-tab loaders) warm that guard concurrently with the read they were making anyway, rather than awaiting a block probe ahead of it.

Discussion comments are filtered **after** the shared comment cache, never inside it — the cache is keyed by document alone, so filtering during the build would let the first reader to open a thread decide what every later reader sees.

## Blocked pages explain themselves

A blocked profile, publication, or article renders the block rather than an empty page or a bare "not found": which direction it runs, and an Unblock control **only** when the block is the reader's own to undo. Someone else's decision isn't theirs to reverse, and a button that pretended otherwise would be a lie.

The article body is withheld outright — `getArticle` returns `null`, and the reason is a separate opt-in read from the not-found state. Nothing about a blocked document, not even its title, crosses the wire on the happy path. A signed-out reader — most of that page's traffic — skips both queries, since only a signed-in reader can be on either side of a block.

**And without a frame of the blocked page first.** Block state rides on the same query as the content it withholds (`getPublicationHeader`, `getAuthorProfile`, `getAuthorSummary`). Resolving it in a second client query would paint the blocked account's hero — name, avatar, description — and then take it away. Mention hovercards read those same queries, so an inline chip can't summarise someone the profile page withholds.

**A block outranks an opt-in the reader made earlier.** The weekly digest, push notifications, and per-reader RSS (`/feed/latest/$did`, `/feed/l/$did/$rkey`) are filtered too. All three arrive uninvited: an emailed article sits in the inbox, a notification lands on a lock screen, and a reader who subscribed in an RSS client can't just not look. "I asked for this source", said before the block, doesn't survive it.

**Blocking is confirmed; failing to block is said out loud.** The overflow-menu item opens a confirmation before writing a public record the whole network honours — the menu offers no way back, because the profile it sits on becomes a notice. Every block mutation reports failure with a toast; silently doing nothing is indistinguishable from having worked.

## Writing needs consent; reading doesn't

`upgradeToBlocking` requests `repo?collection=app.bsky.graph.block&collection=app.bsky.graph.listblock` on the first Block press, persisted via `user.blockingEnabled` so later logins re-request it silently. Same progressive pattern as Margin/Semble. A reader who never blocks from here never sees the prompt; one who signed in before this existed still has every block they made elsewhere enforced.

Block lives in the author overflow menu (last, behind its own separator — it's the one item that hides things rather than opening them). `/settings/blocks` pages through the reader's own blocks and lists their subscribed block lists, saying in words when a list is larger than we mirror.

## Migration

`0035`: `blocks`, `block_lists`, `block_list_items`, `block_list_sync_state`, `block_sync_state`, plus `user.blocking_enabled`. Additive only.

`0036`: drops `blocks_blocker_idx`, a strict prefix of `blocks_edge_idx (blocker_did, subject_did)` — it cost writes to answer nothing the composite couldn't. Kept as a separate migration rather than folded into `0035`, since `0035` may already be applied on a dev branch and rewriting its hash is how you strand a migration ledger.

## What the review pass changed

This describes the final state; `b88790a` is a second pass over the first cut, after an adversarial review. Worth knowing which parts moved, if you read the earlier version:

- The SQL block predicate was applied to **every** signed-in reader, in the one predicate `queries.ts` documents as having caused a 1651ms regression. Now gated behind `blockFilterDid`.
- Incoming blocks were additive-only and **permanently unremovable** in the common case.
- `publicationOwnerDid` was dead — defined, tested, carried by nothing.
- A blocked publication painted in full for a frame before the notice replaced it.
- Blocking a list, and Settings → Refresh, held the HTTP request open for the whole sweep.
- Friend-rail counts and the Discover avatar stack were computed pre-filter, so blocked accounts' avatars still rendered.
- Digest, push and per-reader RSS honoured no blocks at all.

## Verification

`pnpm lint`, `format:check`, `typecheck`, `build`, and `test` (897 passed) all green. 23 unit tests cover the pure block logic — including a regression guard that pins the fixtures to real `ArticleCard` / `PublicationCard` field names, so the dead-`publicationOwnerDid` hole can't silently reopen under a rename.

**Not smoke-tested against live data** — no DB in the authoring environment. Two things worth one run against a real account with blocks before merge:

- The Constellation backlink sources (`app.bsky.graph.block:.subject`, `app.bsky.graph.listitem:.subject`) are the part most likely to need a `forceLegacyLinks` tweak.
- `pnpm perf:explain` against prod-scale data, to confirm the block predicate's plan in practice and not just in principle.

## Known gaps (documented in TODO.md, not papered over)

- **The block UI ships untranslated.** The strings are extracted into all ten catalogs, but the translations are empty — they render in English outside `en` until a translation pass runs.
- **The sweep guard is per-process.** Enough for the single web service today, but a second instance could still sweep the same reader concurrently. Wants an advisory lock.
- **Large blocklists are mirrored partially** (cap ~10k members) and flagged `truncated`. The settings page now says what "Partial" means rather than just badging it.
- **Mutes aren't handled.** `muteActor` / mute lists are a softer signal and out of scope here.
- **Comment badge counts stay unfiltered** for a blocking reader — a per-reader count would need a per-reader cache, and a badge one too high is a smaller wrong than a blocked reply leaking into someone else's thread. The Discussion section now *explains* the gap ("2 more replies are hidden by your blocks") instead of showing "No discussion yet" under a badge reading 6.
- **The collection editor is deliberately unfiltered.** A curator editing their own list should see everything they put in it.

https://claude.ai/code/session_01UY5qvjn48qNmaY3wAoVptX

---
_Generated by [Claude Code](https://claude.ai/code/session_01UY5qvjn48qNmaY3wAoVptX)_

### DIFF
--- a/apps/standard-reader/APP_VISION.md
+++ b/apps/standard-reader/APP_VISION.md
@@ -847,6 +847,15 @@ review-only OAuth client metadata/callback path rather than widening the default
 client metadata scope. Granted scopes may come back either as `include:` sets or
 expanded `repo:` tokens, so permission checks must accept both formats.
 
+**Blocks** are granular too, and one-sided in an unusual way: _reading_ someone's blocks needs
+no scope at all (`com.atproto.repo.listRecords` is public), so a reader's existing blocks are
+enforced from their first page load whatever they granted. Only _writing_ one needs consent — a
+granular `repo?collection=app.bsky.graph.block&collection=app.bsky.graph.listblock` scope,
+requested by `upgradeToBlocking` the first time they press Block and persisted via
+`user.blockingEnabled` so later logins re-request it silently. It targets Bluesky's lexicon
+deliberately: a block is a portable record, and one that only worked in this app would not be a
+block. See "Blocks & block lists" below.
+
 **userinput.app feedback** takes a different approach: `app.userinput.discussion`
 and `app.userinput.upvote` have no permission-set lexicons, so the default OAuth
 client advertises granular `repo?collection=app.userinput.discussion` and
@@ -1167,6 +1176,57 @@ Standard Reader speaks the standard AT Proto label protocol, so readers can subs
   ever needs to be portable, a record + mirror can be layered on the same table.
   `document_push_queue` is the send outbox. **Notifications are independent of subscribing** —
   the bell doesn't change your feed and doesn't require a subscription.
+
+### Blocks & block lists
+
+A reader's blocks follow them here, in both directions, with no setup and no import step.
+
+- **A block is a record, not a setting.** `app.bsky.graph.block` already lives in the blocker's
+  repo and every AT Protocol app that honours a block reads that same record. So there is no
+  `app.standard-reader.block` and there never should be: we mirror Bluesky's lexicon, and blocking
+  from here writes the record the rest of the network already reads. Block someone in Standard
+  Reader and they are blocked on Bluesky too, and wherever the reader goes next. This is the
+  opposite call from labeler subscriptions, and for a concrete reason — Bluesky keeps _those_ in a
+  preferences blob with no per-key write, which is why they had to be ported into records of our
+  own.
+- **Honouring a block costs no OAuth scope.** `com.atproto.repo.listRecords` is public, so the
+  reader's existing blocks are read straight from their PDS and enforced from their first page
+  load. Only _writing_ a block needs consent (`BSKY_BLOCK_WRITE_SCOPE`, persisted via
+  `user.blockingEnabled`), and it is requested the first time they press Block rather than on
+  sign-in — a reader who never blocks from here never sees the prompt, and one who signed in before
+  the feature existed still has every block they made elsewhere enforced.
+- **Four ways to be blocked, all equal in effect.** The reader blocks someone; someone blocks the
+  reader; the account is on a moderation list the reader blocks (`app.bsky.graph.listblock` +
+  `listitem`); or the account blocks a list the reader is _on_ (Bluesky's `blockedByList`). All
+  four hide content, which is what makes a block feel like a block rather than a mute.
+- **Blocks can't ride the tap.** The firehose ingester tracks the repos we index; a block against
+  one of our readers can be written by any repo on the network. So blocks are pulled **per reader,
+  on a cadence** (`server/blocks/sync.server.ts`): their own records from their PDS, incoming
+  blocks from Constellation's backlink index, and list membership from the public AppView. The
+  sweep is scheduled from the OAuth callback and the signed-in shell read, TTL-guarded so a page
+  load between sweeps costs nothing, and forced inline right after a block/unblock write so the
+  change takes effect on the very next page.
+- **Outgoing replaces, incoming only adds.** The reader's own records are the authoritative set, so
+  a sweep can soft-delete the ones that are gone. Blocks _against_ them come from a best-effort
+  index, and a sweep that came back short must never read as "nobody blocks you any more" — that
+  would un-hide content the moment Constellation hiccuped.
+- **Enforced in SQL where a page paginates, in JS where it doesn't.** Feeds, search, directories and
+  archives filter with a `NOT EXISTS` predicate (`notBlockedByViewer`) so a block never shortens a
+  page or strands results behind an offset; rails, profile tabs and third-party discussion
+  (Bluesky, Margin, Semble, pckt, Leaflet) post-filter, where a shorter list is fine. Every helper
+  is bounded by the DIDs a page is about to render, never by the reader's block set — a reader on a
+  40k-member blocklist has to stay an index probe.
+- **Blocked pages explain themselves.** A blocked profile, publication or article renders the block
+  rather than an empty page or a bare "not found": which direction it runs, and an Unblock control
+  only when the block is the reader's own to undo. The article body is withheld outright — the
+  reason is a separate, opt-in read, so nothing about a blocked document (not even its title)
+  crosses the wire.
+- **Big blocklists are mirrored partially, and say so.** The largest public lists run to six
+  figures; past the mirror cap a list is stored truncated and flagged, because silently
+  under-enforcing a blocklist is worse than admitting the limit.
+- **Settings → Blocked accounts** lists the reader's own blocks and their subscribed block lists,
+  with how much of each list is mirrored. It only offers Unblock for blocks they made: someone
+  else's decision is not theirs to undo, and a button that pretended otherwise would be a lie.
 
 ---
 
--- a/apps/standard-reader/APP_VISION.md
+++ b/apps/standard-reader/APP_VISION.md
@@ -1206,27 +1206,64 @@ A reader's blocks follow them here, in both directions, with no setup and no imp
   sweep is scheduled from the OAuth callback and the signed-in shell read, TTL-guarded so a page
   load between sweeps costs nothing, and forced inline right after a block/unblock write so the
   change takes effect on the very next page.
-- **Outgoing replaces, incoming only adds.** The reader's own records are the authoritative set, so
-  a sweep can soft-delete the ones that are gone. Blocks _against_ them come from a best-effort
-  index, and a sweep that came back short must never read as "nobody blocks you any more" — that
-  would un-hide content the moment Constellation hiccuped.
+- **Outgoing replaces; incoming replaces only when the sweep was complete.** The reader's own
+  records are authoritative, so a sweep soft-deletes the ones that are gone. Blocks _against_ them
+  come from a best-effort index, so the delete is gated on the sweep having read every page: a
+  truncated or failed sweep adds but never removes, because it must not read as "nobody blocks you
+  any more". It does have to remove eventually, though — the blocker is usually not a reader here,
+  so nothing else would ever retire the row, and an unblock they made on Bluesky would otherwise
+  hide that account from the reader permanently.
 - **Enforced in SQL where a page paginates, in JS where it doesn't.** Feeds, search, directories and
   archives filter with a `NOT EXISTS` predicate (`notBlockedByViewer`) so a block never shortens a
   page or strands results behind an offset; rails, profile tabs and third-party discussion
-  (Bluesky, Margin, Semble, pckt, Leaflet) post-filter, where a shorter list is fine. Every helper
-  is bounded by the DIDs a page is about to render, never by the reader's block set — a reader on a
-  40k-member blocklist has to stay an index probe.
-- **Blocked pages explain themselves.** A blocked profile, publication or article renders the block
-  rather than an empty page or a bare "not found": which direction it runs, and an Unblock control
-  only when the block is the reader's own to undo. The article body is withheld outright — the
-  reason is a separate, opt-in read, so nothing about a blocked document (not even its title)
-  crosses the wire.
+  (Bluesky, Margin, Semble, pckt, Leaflet) post-filter, where a shorter list is fine. Both check the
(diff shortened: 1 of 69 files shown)


hipstersmoothie:
`standard-reader - recompute-cron` is red on `2af5a81`. I can't attribute it to this diff, and I can't confirm the root cause from here — details below so someone with Railway access can settle it quickly.

**It's the scheduled run, not the deploy.** Deployment `e627e36e` reported ✅ Success at 06:06 alongside every other service. The same deployment ID flipped to 🚨 Crashed at **07:09** — the top of the hour, which is `recompute-cron`'s `0 * * * *` schedule plus Railway jitter. The preview env was created at ~06:05, so 07:00 was its first scheduled execution. All 7 other Railway services are green, and every GitHub Actions check passes.

**Why I don't think it's this branch:**

- `git diff origin/main...HEAD` touches **zero** files under `src/server/ingest/` or `scripts/`. `recompute-cron.mjs` and `recompute.ts` are untouched.
- The one transitive link is `recompute.ts` → `series.ts` → `queries.ts` → my new `server/blocks/blocks.ts`. I suspected that module's top-level `react` `cache()` call might throw in the ingest worker's plain-Node runtime, which would have been mine to fix — but loading both modules under `tsx` with no React context succeeds, and `notBlockedByViewer` builds fine. Disproven.
- Migration `0035` is additive only (five new tables + a nullable `user.blocking_enabled`). It can't break an existing recompute query whether or not it has been applied, and `web`'s `preDeployCommand` ran it successfully at 06:08.

**What I can't do:** the Railway CLI isn't installed in this environment and I have no project credentials, so I can't pull the crash logs for `e627e36e`. `recompute-cron.mjs` throws on any non-2xx from `POST ingest.railway.internal:3099/api/ingest/recompute`, or if the fetch itself fails — so the log line will say which, and that distinguishes "ingest unreachable in an ephemeral preview env" from "recompute genuinely errored".

Worth noting for comparison: on #147 the `recompute-cron` status was posted at deploy time and that PR merged well before its preview env saw a top-of-hour run, so a green status there isn't evidence this is new.

If the logs show a real recompute failure caused by this branch, I'll fix it — happy to pick it up with a paste of the output.

---

hipstersmoothie:
Update on the `recompute-cron` red above: it's green again on `b88790a` — no logs needed, don't go digging.

But green here means the **deploy** of the new `66e30da9` succeeded, which is not the same thing as the scheduled run passing. The crash I reported was the `0 * * * *` execution, and the previous deploy was also green before its first top-of-hour run crashed it. The real test is the ~08:00 run. I'll re-check then and follow up only if it crashes again.

---

hipstersmoothie:
Correcting my earlier comment, and reporting a second failure.

## I was wrong that this diff can't be the cause

I said the diff "touches zero ingest/scripts files." I based that on `git diff origin/main...HEAD -- '*scripts/*' '*ingest*'`, which returns empty — but that glob was too narrow to support the conclusion I drew from it. The diff **does** change code the crons execute:

- `src/server/digest/builder.ts` — what `digest-cron` runs
- `src/server/push/fan-out.ts` — what `push-cron` runs

So "unrelated to this PR" was not a supported claim for the cron failures generally. Please don't rely on my earlier framing.

## `digest-cron` is now also crashing

`digest-cron` succeeded on deploy at 07:46 and then **crashed at 14:02** — its first scheduled run. `builder.ts` is changed here: it now calls `filterBlockedCards` over all four digest sections. That is a plausible cause and I'd treat it as suspected-this-PR until logs say otherwise.

## `recompute-cron`: reproducible hourly, five hours running

Deploy succeeds every time; the scheduled run crashes every time, ~8–12 min past the hour:

| Deploy | Scheduled run |
| --- | --- |
| 07:44 ✅ | 08:08 🚨 |
| 09:05 ✅ | 09:11 🚨 |
| 10:05 ✅ | 10:11 🚨 |
| 12:02 ✅ | 12:08 🚨 |
| 13:06 ✅ | 13:12 🚨 |
| 14:05 ✅ | 14:11 🚨 |
| 15:05 ✅ | 15:12 🚨 |

`scripts/recompute-cron.mjs` only POSTs `ingest.railway.internal:3099/api/ingest/recompute` and throws on a non-2xx, so the crash means ingest returned an error or the request failed — the work itself happens in `ingest`, whose source this PR does not touch. That's weaker evidence than I previously implied, though: I can show the diff doesn't change ingest's code, not that ingest's behaviour is unaffected.

## Hypothesis I tested and disproved

I suspected the module-scope `cache()` from `react` in `src/server/blocks/blocks.ts` (`readerHasBlocks`) would throw in the plain-Node cron runtime, which would hit the digest path. It does not: calling a `cache()`-wrapped async fn outside any React request context returns normally on react `19.3.0-canary`. Ruled out — this time including the call, not just the module load, which is all my earlier check covered.

## What I can't do

There's no `railway` CLI and no Railway credentials in this environment, so I cannot read the crash logs and cannot name the actual error. Someone with dashboard access checking these two would settle it in a minute:

- `digest-cron` deploy `1fc96032-7793-4849-bcc8-60bd4344c31c`
- `recompute-cron` deploy `0e902ab7-e58d-4489-854f-48327cb1113d`

If the digest stack trace points into `filterBlockedCards` or the `blocks*` tables, it's mine and I'll fix it. Also worth confirming whether `recompute-cron` is red on `main`/production too — if it is, that separates it from this branch cleanly, which is the check I should have proposed instead of the PR #147 baseline I already retracted.

---

hipstersmoothie:
Narrowing the `digest-cron` crash — and softening my "suspected-this-PR" call from the previous comment.

Traced the run order in `src/server/digest/run.ts`. Two independent things can throw, and mine is only the first:

1. **`buildDigestForUser`** (run.ts:86) → `filterBlockedCards` over all four sections. **This PR's change.**
2. **`sendEmail`** → `comail.ts` reads `digestConfig.comailFrom` / `comailDid` / `comailApiKey`, and each is a `required()` getter in `config.ts` that **throws `"<NAME> is not set"` when the env var is missing**. **Pre-existing — unrelated to blocks.**

Step 2 runs after step 1, so both are reachable, but step 2 is exactly what a preview environment would trip over: you would not hand a PR preview env live mail-sending credentials, and nothing about that failure needs this branch. If that's it, `digest-cron` is red on every preview environment, and has been since long before this PR.

That makes the log read a one-liner:

- `Error: COMAIL_API_KEY is not set` (or `COMAIL_DID` / `COMAIL_FROM`) → **pre-existing preview-env config gap, not this PR.** Worth fixing separately, or accepting that `digest-cron` is permanently red in previews.
- A stack frame in `filterBlockedCards` / `blockedDidsAmong` / any `blocks*` relation → **mine, and I'll fix it.**

Two hypotheses I've now checked and ruled out, so nobody re-treads them:

- `schema.blocks` being undefined in the cron's schema object — `db/schema.ts` re-exports `./schema/blocks.ts`, all five tables present.
- React's module-scope `cache()` throwing outside a request context — verified it returns normally on react `19.3.0-canary`.

`recompute-cron` is unaffected by any of this and still crashes on every hourly run; its script only POSTs `ingest`, whose source this PR doesn't touch.

---

PR #155, opened 2026-08-12
Resume where you left off (+ Continue reading on /history)

hipstersmoothie:
Reopening an article puts you back where you stopped, and `/history` opens with a shelf of what you're partway through.

## Why this isn't a record

Every other piece of personal state in the app is an AT Proto record in the reader's repo. This one deliberately isn't.

Reads and bookmarks being public is a fair trade — they say *what* you read. A position says how far you got and where you gave up, which is a running commentary on attention that nobody opts into by opening an article. So it lives in a private app-local Neon table (`reading_progress`, keyed `(owner_did, document_uri)`) that the tap never writes, mirrored to `localStorage` per device.

## Two stores, on purpose

They answer different questions:

- **`localStorage`** can be read *before first paint*, which is the only place a scroll restore can happen without a visible jump. It's also what keeps this working offline in the installed PWA.
- **The server row** carries position from phone to laptop. It only wins when it's newer than the local copy *and* the reader hasn't already started scrolling.

## The settle window

Restoring by fraction assumes the article is as tall as it was last time, and for a few hundred milliseconds it isn't — images haven't reserved space yet, so the target offset drifts as they load. The target is re-applied across a 3s window, ended immediately by any real input (`wheel`, `touchstart`, `keydown`, `pointerdown`). Nothing yanks the page out from under someone who has started reading.

A pill names the jump and offers "Start from top", because landing mid-article with no explanation reads as a bug rather than a feature.

## Invariants

Rows exist only between 2% and 95% — below that you didn't start, above it you finished and holding you at 97% forever means the shelf never empties. Writes enforce the band by deleting outside it, so **the table *is* the shelf** and never needs sweeping.

Clearing reading history clears positions (server rows and the local mirror). The whole feature is gated on the existing **Track reading history** setting: position is reading history at a finer grain, and someone who turned that off didn't ask for a more detailed version of it.

## Offline

- Position writes use `networkMode: "offlineFirst"` + `retry: false` rather than React Query's default. The default *pauses* mutations and replays them on reconnect — scrolling emits one write per second, so a subway ride would queue dozens that all replay to say what the last one says. The local copy is the durable record; a single flush on the `online` event sends the value that matters.
- "Start from top" *keeps* the default network mode — it's one deliberate action, not a stream, and dropping it would let the server position come back and undo what the reader asked for.
- The shelf query is **non-fatal** on `/history`: caught in the loader and a plain `useQuery` (not `useSuspenseQuery`) in the component. Offline it doesn't render, rather than taking the reader's cached history down with it.
- The shelf reuses `ArticleRow`, so it inherits the existing offline dimming for articles whose bodies aren't on device.

## Also

Replaces the ad-hoc scroll effect in `article-view.tsx` — the new hook owns the sticky-chrome progress bar it used to compute.

## Before deploy

`pnpm db:migrate` (migration `0037`) — the shelf query 500s without the table.

## Testing

- 12 new unit tests over the local store (band enforcement, LRU eviction at the 100-entry cap, corrupt-value recovery).
- Full suite: 925 passed / 9 skipped. `pnpm lint`, `pnpm typecheck`, `pnpm build` all clean.
- **Not verified in a browser.** The worktree has no `.env` and migration `0037` isn't applied to any database yet, so there was nothing to run against — and scroll-position behaviour is the case where driving Safari from a script reliably lies. Worth a manual pass on: the restore landing point on an image-heavy article, and the cross-device correction.

## Known gaps (deliberate)

- `/history` still isn't in `OFFLINE_ROUTES`, so the shelf is unavailable offline. It degrades to absent, never to an error. Expanding offline route coverage felt out of scope here.
- An offline "Start from top" is a paused mutation, so killing the PWA before reconnect leaves the server row and the position comes back. Same durability level as `markRead`/bookmark today; both want the write outbox the app doesn't have yet.

### DIFF
--- a/apps/standard-reader/APP_VISION.md
+++ b/apps/standard-reader/APP_VISION.md
@@ -576,6 +576,10 @@ comic issue with no pages falls back to it outright.
 
 - Signed-in reader's **reading history** (`app.standard-reader.read`), newest first — every
   article opened while signed in.
+- Opens with a **Continue reading** shelf: up to six articles the reader is partway through,
+  most recently touched first, each with a percentage-read meter. Short on purpose — it is a
+  prompt to pick something back up, not another backlog. It is an extra on this page, never
+  fatal: offline (or on any error) it simply doesn't render and the cached history still does.
 - Route `/history`; linked from the user menu. Requires auth (redirects to login).
 
 ### Reader profile (saved for later)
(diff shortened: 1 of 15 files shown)


hipstersmoothie:
Superseded by #156, which contains this work plus offline coverage for /history and the scroll-restore fixes.

PR #160, opened 2026-08-14
Mute publications and users

hipstersmoothie:
## What

A reader can now **mute a publication or a user**. The muted subject disappears from feeds and discovery — home, latest, discover, trending, tag, search, and the recommendation rails — while **direct navigation still works and subscriptions stay intact**. Muting a user hides everything they touch: documents they authored, documents from publications they own, and their recommends. One direction only; nobody is notified.

Backed by one new repo record, **`app.standard-reader.graph.mute`**: `subject` is a format-less string holding either a user DID or a `site.standard.publication` AT-URI (the lexicon has no did-or-at-uri format; app code classifies), keyed by the deterministic `subjectRkey` so mute ⇄ unmute address one record. Note the record is public in the reader's repo — the *effect* is private, the record is not (unlike Bluesky's server-side mutes).

## How

- **Lexicon + scope** — `lexicons/app/standard-reader/graph.mute.json` (reuses the published `graph` NSID authority, no DNS work) and the collection added to the `authBasicFeatures` permission set. `hasMuteWriteScope` accepts the verbatim `include:` token or a PDS-expanded `repo` token; the write path additionally maps a PDS 401/403 to `needs-mute-scope` → `upgradeToMuting` re-auth, covering sessions whose grant predates the set republish (plain basic scope, no new consent tier or flag).
- **Read-model** — `mutes` table (**migration `0039`**, rebased cleanly on top of #159's `0038_weekly_thread_runs`; `drizzle-kit check` passes) with discriminated `subject_did` / `subject_uri` columns and muter/edge indexes. Firehose ingest (`upsertMute` + delete case), repo-sync repair registration, tap collection filters. Writes eagerly mirror the row so a mute takes effect on the very next page.
- **Enforcement** — new `src/server/mutes/mutes.ts`, mirroring the blocks architecture and its rules (no `.server` suffix, bounded by the candidate set, never the mute set):
  - `muteFilterDid` has-any gate (reactCache + 60s TTL, invalidated on write) — readers who mute nothing pay zero; the only supported route to the predicate.
  - `notMutedByViewer` — correlated `NOT (EXISTS …)` with author-DID, publication-owner-DID, and a publication-URI branch blocks don't have; each an index probe on the edge indexes. Threaded as a parallel `muterDid` option through `queries.ts` (block call sites untouched).
  - `filterMutedCards` post-filter on rails; the home/latest follow feeds also **pre-filter muted sources out of their input sets**, which is what removes a muted user's recommend-sourced entries and keeps the Latest tab badge counts agreeing with the rows.
  - Deliberately **not** applied to publication/author/article pages, curated lists, or the subscriptions table — muting is scoped to feeds + discovery.
- **UI** — `MuteMenuItem` + `MuteDialog` (MenuItem/AlertDialog as siblings, per the `BlockUserMenuItem` doc) in the author and publication overflow menus next to Block; mute confirms, unmute is immediate. New **Settings → Muted** page listing both kinds with unmute (rows still link to the pages they hide), plus a Muted row in the settings index. Strings extracted into all ten Lingui catalogs.
- **Tests** — `src/lib/mutes.test.ts` (subject classification + card-rejection matrix), `scope.test.ts` (`hasMuteWriteScope` token forms), and a `muterDid` variant in the gated EXPLAIN perf suite (`pnpm perf:explain`).

## Verification

- `pnpm typecheck`, `pnpm lint`, `pnpm format:check`, `pnpm lex:lint`, `pnpm build`, `drizzle-kit check` all pass; 41 unit tests pass.
- **Perf verified against prod-scale data** on a throwaway Neon branch (migration applied + mutes seeded for the heaviest perf reader): the mute-filter EXPLAIN variants keep the follow-feed candidate query on its per-source index scans — no `Seq Scan on documents`, exec time within the same budget as the block filter, alone and with both filters stacked. The two pre-existing perf-spec budget failures reproduce identically from a clean `origin/main` checkout on the same database (not from this PR).
- Not yet done: a live dev pass through the UI.

## ⚠️ Deploy ordering

1. **Publish before deploy**: `pnpm atproto:publish-lexicons` must push the new `graph.mute` schema record **and** the republished `authBasicFeatures` set before this deploys, or fresh logins won't be granted the collection.
2. **Tap filters**: pre-Jetstream deploys need `app.standard-reader.graph.mute` appended to the live `TAP_COLLECTION_FILTERS` env override on Railway (compose defaults updated here; becomes a no-op once #158's `app.standard-reader.*` wildcard lands).
3. **Jetstream branch (#158)**: still claims migration `0038` against the same parent — whichever of it and `main` merges second now regenerates (this PR is already renumbered to `0039` past #159's `0038`).
4. Pre-republish sessions may see one extra consent screen on their first mute (expected; the dual scope check handles it).

Follow-ups tracked in TODO.md: scrub "Recommended by X" bylines on cards surviving via a non-muted source; translation pass; digest/push/RSS parity with blocks.

### DIFF
--- a/apps/standard-reader/APP_VISION.md
+++ b/apps/standard-reader/APP_VISION.md
@@ -1366,6 +1366,11 @@ reversibly.
 - **UI**: Mute/Unmute in the author and publication overflow menus (next to Block; mute confirms,
   unmute is immediate), and **Settings → Muted** lists both kinds with unmute — every row still
   links to the page it hides, because muting never takes those pages away.
+- **A muted page says so.** Because muting leaves the profile readable, a muted author or
+  publication page would otherwise look identical to an unmuted one. A "Muted" pill renders on the
+  header in the same visual register as a labeler label (`MutedPill`, next to `AccountLabels`) —
+  it's the reader's own statement about the account — with a hover card explaining the scope and
+  carrying the Unmute action and a link to Settings → Muted.
 
 ---
 
--- a/apps/standard-reader/APP_VISION.md
+++ b/apps/standard-reader/APP_VISION.md
@@ -892,7 +892,7 @@ upstream `site.standard.auth*` sets (published by standard.site — see
 
 | Tier                                | App-owned set (we publish)                      | site.standard set (we reference)         | Covers                                                                               |
 | ----------------------------------- | ----------------------------------------------- | ---------------------------------------- | ------------------------------------------------------------------------------------ |
-| **Basic** (default sign-in)         | `include:app.standard-reader.authBasicFeatures` | `include:site.standard.authSocial`       | bookmark, read, list, listSave, labelerSubscription + follows + likes                |
+| **Basic** (default sign-in)         | `include:app.standard-reader.authBasicFeatures` | `include:site.standard.authSocial`       | bookmark, read, list, listSave, labelerSubscription, graph.mute + follows + likes    |
 | **Collections authoring** (upgrade) | `+ include:app.standard-reader.authCollections` | swap to `include:site.standard.authFull` | collection, collectionsPublication, publicationTheme + publication + document writes |
 | **Subscribe embed**                 | —                                               | `include:site.standard.authSocial`       | subscription write (also covers recommend)                                           |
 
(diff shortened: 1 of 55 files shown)


hipstersmoothie:
Recreated as a fresh PR (same branch) after the rebase past #159 so the migration state is unambiguous: the mutes migration is now `0039_glamorous_martin_li`, chained cleanly off `0038_weekly_thread_runs`, and `drizzle-kit check` passes. Superseded by the new PR.

PR #167, opened 2026-08-20
Stop the OAuth refresh lock starving the request pool

hipstersmoothie:
## The symptom

A reader reported the site "doesn't load"; it wasn't reproducible from another account.

Their requests weren't erroring — they were **hanging**. `user.getShellBootstrap` returned HTTP 200 with `ok: true` after up to **50 minutes**. Because nothing ever errored, this was invisible in error-rate dashboards.

It wasn't user-specific either. Across all signed-in traffic:

| | P50 | P95 | MAX |
|---|---|---|---|
| all `getShellBootstrap` (mostly guests) | 0.32ms | 1.14ms | — |
| **signed-in only** (4,308 calls / 7d) | — | **584,200ms (9.7 min)** | — |
| the reporting account (24h) | 210,678ms | — | 3,010,486ms |

Guests are fine because the guest path returns before touching the DB. Signed-in readers all share one tail.

## Root cause: pool starvation, not slow SQL

`withAdvisoryLock` holds its transaction — and therefore its pool connection — open across `fn()`, which is the OAuth **token-refresh round trip to the reader's PDS**: a network call to an arbitrary third-party server that atcute only bounds at 30s.

Run against the shared pool (`max: 16`), a handful of slow PDSes drain it. And the pool had **no `connectionTimeoutMillis`**, so node-postgres queued acquires *forever*. Every unrelated signed-in read parked behind those refreshes with no error, no timeout, and then completed with a 200 once a connection freed.

The tell is queries that *cannot* be slow degrading in lockstep with everything else, at the same wall-clock moments:

| span | P50 | P95 |
|---|---|---|
| `sidebarPref.get` — a single-row indexed lookup | 4.7s | **314s** |
| `shell.getShellSnapshot` | 32s | 733s |
| `lists.getAuthorLists` | 21s | 925s |
| `lists.getLists` | 31s | 493s |

Synchronized spikes across unrelated queries = contention on a shared resource. Guest traffic over the same window (77k spans, never touches the DB) stayed at P50 0.65ms / P95 334ms.

## The fix

**Give the advisory lock its own small pool** (`getLockDb`, `max: 8`). A slow PDS can now stall refreshes without touching the 16 connections ordinary reads depend on. Created lazily, so nothing opens a second pool under neon-http or in tests.

The locking semantics are **deliberately untouched**. The obvious-looking fix — moving `fn()` outside the transaction — would drop the cross-process mutual exclusion this file exists to provide and reintroduce the `invalid_grant` logout race that motivated it.

**Require `connectionTimeoutMillis` on every pool.** Unset, exhaustion degrades into silent multi-minute hangs that still answer 200. Set, it fails fast and becomes something a dashboard can see.

New env vars, all with working defaults (documented in `.env.example`): `DB_POOL_ACQUIRE_TIMEOUT_MS` (5000), `DB_LOCK_POOL_MAX` (8), `DB_LOCK_ACQUIRE_TIMEOUT_MS` (20000). `DB_POOL_MAX` (16) is unchanged.

## This is a mitigation, not the complete fix

The refresh still pins a connection across a network round trip — the blast radius is just contained. The real fix is a lease/claim row with a TTL replacing `pg_advisory_xact_lock`, so no connection is held across PDS I/O. That needs a migration, so it's filed in TODO.md rather than bundled here.

## Two unrelated problems this turned up

Both are diagnosis only — no code here addresses them. Filed in TODO.md.

**1. `reconcile-cron` is running a stale build; the sweep is 100% dead.** `ingest.repoReconcile` fails ~99.9% of the time — **43,495 failures over 7d, flat at ~350–440/hr** for the entire window. 96.5% are one query selecting `tracked_repos.last_seen_rev`, the column migration `0041_tidy_last_seen_seq` **dropped**. No current source generates that query, so the deployed service predates 0041; only the web service runs `preDeployCommand` migrations, so the shared DB moved on without it. **No repo has been reconciled for at least a week.** Needs a redeploy, not a patch.

**2. Readers get stuck in a zombie signed-in state.** Nothing reconciles our app `session` row with atcute's stored OAuth session — the only place a session row is deleted is the explicit `user.signOut`. Once the OAuth blob is gone the session cookie stays valid, so the app keeps believing the reader is signed in while silently serving them guest data, forever. `auth.oauth.session_deleted` then fires **every 60s on the dot**, inside a `GET` span, indefinitely; one account has been looping since at least 2026-08-13. Each attempt also takes the advisory lock and a lock-pool connection.

Worth stating precisely, because the raw number is misleading: 25,331 events over 7d is **21 distinct DIDs**, 4 of them 99.8%, and the count is ~2x inflated (one event per OAuth client kind per attempt). The lock itself is healthy — `refresh_lock_unavailable` and `refresh_lock_timeout` are both **0** over 7d.

## Verification

- **Typecheck**: 10 errors before, 10 after — stash-diffed against baseline. All pre-existing `@bsky/jetstream` module-resolution failures from #158; none in changed files.
- **oxlint / oxfmt**: clean.
- **Tests**: `src/integrations/auth` + `src/db` — 6 passed.

Not load-tested against a genuinely saturated pool; the reasoning rests on the telemetry above.

### DIFF
--- a/apps/standard-reader/.env.example
+++ b/apps/standard-reader/.env.example
@@ -16,6 +16,17 @@ DATABASE_URL="postgresql://[user]:[password]@[neon_hostname]/[dbname]?sslmode=re
 # queries skip per-request TLS/proxy overhead. In prod (Railway ↔ Neon, same
 # region) the TCP+TLS handshake is <5ms — far cheaper than per-query HTTP.
 DB_DRIVER="pg"
+# Request pool size, and how long a query may wait for a free connection before
+# it errors. node-postgres queues acquires forever if unset, which turns pool
+# exhaustion into multi-minute hangs that still answer 200 — set this so a
+# drained pool fails loudly instead of silently.
+DB_POOL_MAX="16"
+DB_POOL_ACQUIRE_TIMEOUT_MS="5000"
+# Separate pool for the OAuth refresh advisory lock, which holds its connection
+# across the PDS token-refresh round trip. Isolating it means a slow third-party
+# PDS can stall refreshes without starving ordinary signed-in reads.
+DB_LOCK_POOL_MAX="8"
+DB_LOCK_ACQUIRE_TIMEOUT_MS="20000"
 
 # ──────────────────────────────────────────────────────────────────────────────
 # Analytics — Plausible (client-side, `@plausible-analytics/tracker`)
--- a/apps/standard-reader/TODO.md
+++ b/apps/standard-reader/TODO.md
@@ -503,16 +503,33 @@ stores in `src/integrations/auth/`, session/user server fns in
       transaction to claim, release the connection, refresh, short transaction to
       release) with a TTL so a crashed process can't strand the lease — replacing
       `pg_advisory_xact_lock`. Needs a migration.
-- [ ] **The refresh lock is not actually preventing logouts.** `attachSessionEventLogging` calls
-      `auth.oauth.session_deleted` "the primary breadcrumb for the logout-on-deploy bug" and says a
-      recurring stream means refresh rotation is still racing. Over 7d there are **25,331** of them,
-      and **25,286 (99.8%) carry `cause: "session was deleted by another process"`** — the exact
-      racing-refresh signature, not genuine PDS revocation (that accounts for ~40). Meanwhile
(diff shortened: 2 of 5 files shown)


hipstersmoothie:
Folded into #168 so both fixes ship as one PR — that branch already contains these commits.

PR #170, opened 2026-08-20
Fix two ways a signed-in reader silently loses the site

hipstersmoothie:
Started from a report that the site "doesn't load" for one reader, which wasn't reproducible from another account. It turned out to be two independent bugs, both of which strand signed-in readers while reporting success — so neither showed up in error rates.

---

# 1. The OAuth refresh lock starves the request pool

## The symptom

Requests weren't erroring, they were **hanging**. `user.getShellBootstrap` returned HTTP 200 with `ok: true` after up to **50 minutes**.

It wasn't user-specific either. Across all signed-in traffic:

| | P50 | P95 | MAX |
|---|---|---|---|
| all `getShellBootstrap` (mostly guests) | 0.32ms | 1.14ms | — |
| **signed-in only** (4,308 calls / 7d) | — | **584,200ms (9.7 min)** | — |
| the reporting account (24h) | 210,678ms | — | 3,010,486ms |

Guests are fine because the guest path returns before touching the DB. Signed-in readers all share one tail.

## Root cause: pool starvation, not slow SQL

`withAdvisoryLock` holds its transaction — and therefore its pool connection — open across `fn()`, which is the OAuth **token-refresh round trip to the reader's PDS**: a network call to an arbitrary third-party server that atcute only bounds at 30s.

Run against the shared pool (`max: 16`), a handful of slow PDSes drain it. And the pool had **no `connectionTimeoutMillis`**, so node-postgres queued acquires *forever*. Every unrelated signed-in read parked behind those refreshes with no error, no timeout, and then completed with a 200 once a connection freed.

The tell is queries that *cannot* be slow degrading in lockstep with everything else, at the same wall-clock moments:

| span | P50 | P95 |
|---|---|---|
| `sidebarPref.get` — a single-row indexed lookup | 4.7s | **314s** |
| `shell.getShellSnapshot` | 32s | 733s |
| `lists.getAuthorLists` | 21s | 925s |
| `lists.getLists` | 31s | 493s |

Synchronized spikes across unrelated queries = contention on a shared resource. Guest traffic over the same window (77k spans, never touches the DB) stayed at P50 0.65ms / P95 334ms.

## The fix

**Give the advisory lock its own small pool** (`getLockDb`, `max: 8`). A slow PDS can now stall refreshes without touching the 16 connections ordinary reads depend on. Created lazily, so nothing opens a second pool under neon-http or in tests.

The locking semantics are **deliberately untouched**. The obvious-looking fix — moving `fn()` outside the transaction — would drop the cross-process mutual exclusion this file exists to provide and reintroduce the `invalid_grant` logout race that motivated it.

**Require `connectionTimeoutMillis` on every pool.** Unset, exhaustion degrades into silent multi-minute hangs that still answer 200. Set, it fails fast and becomes something a dashboard can see.

New env vars, all with working defaults (documented in `.env.example`): `DB_POOL_ACQUIRE_TIMEOUT_MS` (5000), `DB_LOCK_POOL_MAX` (8), `DB_LOCK_ACQUIRE_TIMEOUT_MS` (20000). `DB_POOL_MAX` (16) is unchanged.

**This is a mitigation, not the complete fix.** The refresh still pins a connection across a network round trip — the blast radius is just contained. The real fix is a lease/claim row with a TTL replacing `pg_advisory_xact_lock`, so no connection is held across PDS I/O. That needs a migration, so it's filed in TODO.md rather than bundled here.

---

# 2. Readers get stuck in a zombie signed-in state

## The bug

Nothing reconciles our app `session` row with atcute's stored OAuth session. The only place a session row is deleted is the explicit `user.signOut` (`api-user.functions.ts:1568`); `attachSessionEventLogging` only *logs* the `deleted` event.

So once the OAuth blob vanishes the two halves disagree **forever**: the session cookie stays valid, the app keeps believing the reader is signed in, and every request quietly falls back to guest data while re-attempting a restore that can never succeed. The only escape is clearing cookies.

The telemetry signature is unmistakable — `auth.oauth.session_deleted` for one DID:

```
14:35:30  14:36:29  14:37:29  14:38:29  14:39:29
```

Every 60 seconds on the dot, inside a `GET` span, indefinitely. One account has been looping since at least 2026-08-13. Each attempt also takes the advisory lock and a lock-pool connection, feeding bug #1 above.

Scale, measured rather than assumed: 25,331 events over 7d is **21 distinct DIDs**, 4 of them 99.8%.

## Why this was not a one-liner

The obvious fix — tear down the session in the `deleted` listener — is the dangerous version. `restoreAtprotoSession` swallows every error and returns `null`, so **"the session is gone" and "the PDS was briefly unreachable" are indistinguishable to callers**. And `"session was deleted by another process"` can itself fire on a benign race. Acting on that signal from a global, DID-keyed listener would log people out for no reason — the exact failure class the refresh lock was added to fix.

atcute's `deleted` event is the one signal that separates the two cases. So: record it, and consume it at the point a restore actually fails, where the session token is in hand.

Kept deliberately narrow:

- **Fires only on a recorded deletion**, consumed once rather than latched, so a stale record can't sign out a reader who has since signed back in.
- **Ends only the row for the token on this request**, not every session for the DID — one wedged device can't sign the reader out everywhere.
- **Leaves caches and cookie untouched if the row delete fails**, rather than reporting signed-out while the session still works.

The result: a permanent zombie state becomes an ordinary signed-out one the reader recovers from by logging in again.

Tracking lives in its own dependency-free module (`session-deletions.ts`) so the request path can read the signal without importing the OAuth client, and so it's testable without standing up OAuth config.

Also documents something that was quietly skewing the telemetry: `restoreAtprotoSession` tries **both** client kinds and both have listeners, so a single failed restore emits `session_deleted` **twice**. Counts read as roughly double the real number of lost sessions.

---

# A third problem, diagnosis only

No code here addresses it; it needs a redeploy, not a patch. Filed in TODO.md.

**`reconcile-cron` is running a stale build; the sweep is 100% dead.** `ingest.repoReconcile` fails ~99.9% of the time — **43,495 failures over 7d, flat at ~350–440/hr** for the entire window. 96.5% are one query selecting `tracked_repos.last_seen_rev`, the column migration `0041_tidy_last_seen_seq` **dropped**. No current source generates that query, so the deployed service predates 0041; only the web service runs `preDeployCommand` migrations, so the shared DB moved on without it. **No repo has been reconciled for at least a week.**

(Related leftover: `scripts/tmp-drain.ts:19` still filters on `t.last_seen_rev is null` in raw SQL, so it fails at runtime rather than typecheck.)

---

# Verification

Full CI gate run locally, with dependencies installed from the lockfile:

- **`pnpm lint`** — 0 errors, 1,429 files
- **`pnpm format:check`** — clean, 1,609 files
- **`pnpm typecheck`** — clean, **0 errors** across every workspace package
- **`pnpm build`** — succeeds
- **Tests**: 9 new cases in `session-deletions.test.ts` covering the semantics that matter — never firing for an unrecorded DID (the over-fire risk), consuming once across the two client kinds, the TTL boundary, expired records being consumed rather than left to fire later, cross-DID pruning, and re-arming after a re-record. `src/integrations/auth` + `src/db` — 15 passed.

**Not verified:** neither fix is exercised against the real failure conditions — I couldn't reproduce a saturated pool or a wedged account locally, so both rest on production telemetry plus tests of the decision rules. After deploy, worth watching: signed-in `getShellBootstrap` P95 should collapse toward the guest numbers, and `auth.session.ended_after_oauth_delete` should fire a handful of times for those 4 DIDs and then go quiet. If the latter fires steadily instead, the deletion signal is being consumed too eagerly and needs another look.

### DIFF
--- a/apps/standard-reader/.env.example
+++ b/apps/standard-reader/.env.example
@@ -16,6 +16,17 @@ DATABASE_URL="postgresql://[user]:[password]@[neon_hostname]/[dbname]?sslmode=re
 # queries skip per-request TLS/proxy overhead. In prod (Railway ↔ Neon, same
 # region) the TCP+TLS handshake is <5ms — far cheaper than per-query HTTP.
 DB_DRIVER="pg"
+# Request pool size, and how long a query may wait for a free connection before
+# it errors. node-postgres queues acquires forever if unset, which turns pool
+# exhaustion into multi-minute hangs that still answer 200 — set this so a
+# drained pool fails loudly instead of silently.
+DB_POOL_MAX="16"
+DB_POOL_ACQUIRE_TIMEOUT_MS="5000"
+# Separate pool for the OAuth refresh advisory lock, which holds its connection
+# across the PDS token-refresh round trip. Isolating it means a slow third-party
+# PDS can stall refreshes without starving ordinary signed-in reads.
+DB_LOCK_POOL_MAX="8"
+DB_LOCK_ACQUIRE_TIMEOUT_MS="20000"
 
 # ──────────────────────────────────────────────────────────────────────────────
 # Analytics — Plausible (client-side, `@plausible-analytics/tracker`)
--- a/apps/standard-reader/TODO.md
+++ b/apps/standard-reader/TODO.md
@@ -503,16 +503,33 @@ stores in `src/integrations/auth/`, session/user server fns in
       transaction to claim, release the connection, refresh, short transaction to
       release) with a TTL so a crashed process can't strand the lease — replacing
       `pg_advisory_xact_lock`. Needs a migration.
-- [ ] **The refresh lock is not actually preventing logouts.** `attachSessionEventLogging` calls
-      `auth.oauth.session_deleted` "the primary breadcrumb for the logout-on-deploy bug" and says a
-      recurring stream means refresh rotation is still racing. Over 7d there are **25,331** of them,
-      and **25,286 (99.8%) carry `cause: "session was deleted by another process"`** — the exact
-      racing-refresh signature, not genuine PDS revocation (that accounts for ~40). Meanwhile
-      `auth.oauth.refresh_lock_contended` fires only 186×/24h, so the lock is rarely even engaging:
-      something is bypassing `dbRequestLock` rather than the lock failing to hold. Worth checking
-      whether every process that restores a session goes through it (ingest/cron included) and
-      whether `isNeonHttpDriver` is silently true anywhere — that branch degrades to in-process-only
-      serialization and logs `refresh_lock_unavailable` just once per process.
+- [ ] **Readers get stuck in a zombie signed-in state when their OAuth session vanishes.** Nothing
+      reconciles our app `session` row with atcute's stored OAuth session: the only place a session
+      row is deleted is the explicit `user.signOut`
+      ([`api-user.functions.ts:1568`](src/integrations/tanstack-query/api-user.functions.ts)). So
+      once the OAuth blob is gone, the reader's session cookie stays valid, the app keeps believing
+      they're signed in, and every request silently degrades to guest data forever.
+      `attachSessionEventLogging` only *logs* the `deleted` event — it doesn't invalidate anything.
+
+      The telemetry signature: `auth.oauth.session_deleted` fires **every 60 seconds on the dot**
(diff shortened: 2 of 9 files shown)


hipstersmoothie:
Not needed — #168 was already squash-merged to main as e221162a with these exact 9 files. Opened this before noticing the merge.

PR #181, opened 2026-08-29
Make atproto-proxy actually authenticate, and test the XRPC surface end to end

hipstersmoothie:
The `atproto-proxy` interop report ([userinput.app](https://userinput.app/d/did:plc:7qubw2z53qzfturlblaozz4a/3msqg3okp3kc7)) had a third layer under the two we already shipped, and nothing we had would have caught it.

## The bug

`getSigningKey` returned the DID document's bare `publicKeyMultibase` (`zQ3sh…`). `verifyJwt` hands that to `@atproto/crypto`'s `verifySignature`, which parses it as a **did:key** — so every real PDS-minted service JWT failed `BadJwtSignature`. `authenticateRequest` then swallowed the failure and retried the token as an access token, so:

- proxied **queries** silently answered as **anonymous** — `200 {"active":false}`
- proxied **procedures** 401'd with `"Unable to resolve PDS for access token"`

Reproduced against prod with the test account:

```
== the account has this document bookmarked ==
  no auth        200 {"active":false}
  app-password   200 {"active":true}    ← correct
  service JWT    200 {"active":false}   ← silently anonymous
  via proxy      200 {"active":false}   ← silently anonymous
```

`src/server/labeler/verify.server.ts` had always wrapped the key correctly; the XRPC path never did.

## Why it survived two "fixes"

`scripts/verify-atproto-proxy.ts` asserted *"the proxied read returns 200"*. `getBookmarkStatus` is optionally authenticated and answers `{"active":false}` to a stranger, so it returned 200 throughout. **A status code is not an authentication check.**

## The test

`examples/xrpc-client` — a dependency-free external client that imports nothing from the app, logs in with an app password, and runs two things:

**A transport probe** over `direct`, `atproto-proxy` and a self-minted service JWT. It bookmarks a document, then requires each transport to read that bookmark back with no `did` parameter, so the answer has to come from the credential. It also validates the published DID document the way a PDS does — matching `id`, bare-origin `serviceEndpoint` — since both earlier failures lived there.

**A conformance sweep** of all 54 XRPC methods using the example arguments the public API docs advertise. Writes run paired with the procedure that undoes them, `createList`/`updateList`/`deleteList` chain on the returned rkey, and the two procedures with no inverse are skipped unless asked for. Fixtures are discovered at run time from public endpoints, so there is no fixture file to keep current.

```
pnpm --filter @standard-reader/example-xrpc-client start
```

It cannot quietly stop covering something: the plan is generated from `API_DOCS_CATALOG` and a test fails when the checked-in copy drifts, and `registry.test.ts` now also asserts the catalog documents *every* registered method. That last check found five labeler endpoints that had shipped undocumented and untested — they are now in the catalog, the docs page and the sweep.

## A second bug the sweep found

`searchDocuments?q=reader` returned `400 "Internal error"` in production. `documentTierSql` built its author arms as `${did} = any(${authorDids})`; Drizzle expands a JS array in a `sql` template into a parenthesised tuple (`($1, $2)`), which `in` accepts and `any()` rejects as a syntax error. The arms only exist when the profile lookup matched, so it broke exactly the queries matching an author handle or display name — `?q=reader` failed, `?q=atproto` and `?q=readers` were fine.

## Verification

- `pnpm test` — 1185 passed
- `pnpm lint`, `pnpm format:check`, `pnpm typecheck` clean
- 5 new unit tests mint real Secp256k1 service JWTs against a plc-shaped DID document fixture; reverting the one-line fix fails 3 of them
- `scripts/verify-atproto-proxy.ts` and the transport probe both correctly fail against the current (unfixed) production and should go green once this deploys

https://claude.ai/code/session_01FifMWjZHUrN3Nrc2rTLyPB

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -30,6 +30,8 @@ apps/extension/         # the browser extension (WXT); shares app source via #/ 
 packages/               # workspace packages: design-system (shared UI) + the
                         #   publishable @standard-reader/renderer-* family + lexicons
 services/               # standalone labeler services (claudeslop, botlabeler)
+examples/xrpc-client/   # dependency-free external client for the AppView + the
+                        #   end-to-end XRPC conformance run
 scripts/                # workspace-level tooling (package publishing)
 config/oxlint/          # shared lint config — applies to the whole repo
 ```
@@ -37,8 +39,8 @@ config/oxlint/          # shared lint config — applies to the whole repo
 **Reading paths in this file:** bare `src/…`, `scripts/…`, `perf/…`, `drizzle/…`, and
 `lexicons/…` paths are relative to **`apps/standard-reader/`** unless written out in full.
 
-The workspace root (`pnpm-workspace.yaml`) covers `apps/*`, `extension`, `packages/*`, and
-`services/*`. Lint and format run **repo-wide from the root** (one oxlint/oxfmt config); build,
+The workspace root (`pnpm-workspace.yaml`) covers `apps/*`, `examples/*`, `extension`,
+`packages/*`, and `services/*`. Lint and format run **repo-wide from the root** (one oxlint/oxfmt config); build,
 typecheck, test, and every app task are `pnpm --filter` passthroughs — so `pnpm dev`, `pnpm build`,
 `pnpm test`, `pnpm db:migrate`, etc. still work unchanged from the root.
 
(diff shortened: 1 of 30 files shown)


hipstersmoothie:
## Verified against the preview deployment

`https://web-standard-reader-pr-181.up.railway.app` publishes a self-consistent DID document for its own host, so the whole proxy path is exercisable there:

```json
{"id":"did:web:web-standard-reader-pr-181.up.railway.app",
 "service":[{"id":"#standard_reader_appview","serviceEndpoint":"https://web-standard-reader-pr-181.up.railway.app"}]}
```

**Transport probe — all four pass** (they all fail on production today):

```
PASS  direct: reads as the signed-in reader
PASS  proxy: reads as the signed-in reader
PASS  serviceAuth: reads as the signed-in reader
PASS  proxy: explains why a proxied write cannot work
```

**Conformance sweep — 54 methods, 52 passed, 2 skipped, 0 failed.** `searchDocuments` passes here and fails on production, confirming the `= any()` fix against a real database.

`scripts/verify-atproto-proxy.ts` is green against the preview too.

## Two more things the sweep found, now fixed

- `handleFollowUser` threw a plain `Error("You can't follow yourself.")`. `handleXrpcError` will not leak non-XRPCError messages, so callers got `400 InvalidRequest "Internal error"` and learned nothing — it is an `InvalidRequestError` now.
- The API docs' `followUser` example passed `readerDid`, the *calling* reader. The one call a developer is most likely to try from the docs page was the one guaranteed to fail. Split out a `followTargetDid` fixture.

`pnpm test` 1186 passed; lint, format:check, typecheck clean.

